PR #18, opened 2022-05-13
[DRAFT] Enabling auth cookies to be shared between apps

mjrousos:


mjrousos:
Closing this to focus on deferring auth decisions to the ASP.NET app instead. That seems like a more robust solution. Sharing auth cookies still has some advantages and is something we may want to look into for some subset of scenarios in the future.

PR #19, opened 2022-05-16
Remove debug builds

Tratcher:
I don't think anything in the repo is debug/release sensitive so we can remove these redundant builds.


wtgodbe (approved):

twsouthwick:
We don't currently have any #DEBUG kind of stuff right now, so this seems fine

twsouthwick (approved):

PR #20, opened 2022-05-16
[DRAFT] Enable shared auth by forwarding claims principal with proxied requests

mjrousos:


mjrousos:
Closing this to focus on deferring auth decisions to the ASP.NET app instead (more or less the opposite of this approach).

PR #21, opened 2022-05-16
Implement ISessionState against ISession

twsouthwick:
This enables using the in-built ASP.NET Core session through the adapters (i.e. strongly typed). It still requires people to use the adapters to access it, but provides a pathway to remove reliance on the .NET Framework app.

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/BuiltIn/BuiltInSessionExtensions.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/BuiltIn/BuiltInSessionExtensions.cs
@@ -1,0 +1,27 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using Microsoft.AspNetCore.SystemWebAdapters.SessionState.BuiltIn;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.AspNetCore.SystemWebAdapters;
+
+public static class BuiltInSessionExtensions
+{
+    public static ISystemWebAdapterBuilder AddBuiltInSession(this ISystemWebAdapterBuilder builder, Action<Builder.SessionOptions>? options = null)
+    {
+        if (options is null)
+        {
+            builder.Services.AddSession();
+        }
+        else
+        {
+            builder.Services.AddSession(options);
+        }
+
+        builder.Services.AddSingleton<ISessionManager, BuiltInSessionManager>();
+
+        return builder;
+    }
+}
--- a/samples/ClassLibrary/SessionUtils.cs
+++ b/samples/ClassLibrary/SessionUtils.cs
@@ -1,4 +1,4 @@
-using Microsoft.AspNetCore.SystemWebAdapters.SessionState;
+using Microsoft.AspNetCore.SystemWebAdapters.SessionState.Serialization;
 
 namespace ClassLibrary;
 
@@ -6,7 +6,7 @@ public class SessionUtils
 {
     public static string ApiKey = "test-key";
 
-    public static void RegisterSessionKeys(SessionOptions options)
+    public static void RegisterSessionKeys(SessionSerializerOptions options)
     {
         options.RegisterKey<int>("test-value");
         options.RegisterKey<SessionDemoModel>("SampleSessionItem");
--- a/samples/MvcApp/Global.asax.cs
+++ b/samples/MvcApp/Global.asax.cs
@@ -18,11 +18,9 @@ namespace MvcApp
 
             Application.AddSystemWebAdapters()
                 .AddProxySupport(options => options.UseForwardedHeaders = true)
-                .AddRemoteAppSession(options=>
-                {
-                    options.ApiKey = ClassLibrary.SessionUtils.ApiKey;
-                    ClassLibrary.SessionUtils.RegisterSessionKeys(options);
-                });
+                .AddRemoteAppSession(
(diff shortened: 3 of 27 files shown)


src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/BuiltIn/BuiltInSessionExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+
+using System;
+using Microsoft.AspNetCore.SystemWebAdapters.SessionState.BuiltIn;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.AspNetCore.SystemWebAdapters;
+
+public static class BuiltInSessionExtensions
+{
+    public static ISystemWebAdapterBuilder AddBuiltInSession(this ISystemWebAdapterBuilder builder, Action<Builder.SessionOptions>? options = null)
```
Tratcher, line 12: BuiltIn is a non-descript API name. Rename all the BuiltIn types.
```suggestion
    public static ISystemWebAdapterBuilder WrapAspNetCoreSession(this ISystemWebAdapterBuilder builder, Action<Builder.SessionOptions>? options = null)
```

src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/BuiltIn/BuiltInSessionExtensions.cs
```diff
@@ -0,0 +1,27 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using Microsoft.AspNetCore.SystemWebAdapters.SessionState.BuiltIn;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.AspNetCore.SystemWebAdapters;
+
+public static class BuiltInSessionExtensions
```
Tratcher, line 10:
```suggestion
public static class ISystemWebAdapterBuilderSessionExtensions
```

Tratcher (approved):

docs/session-state/remote-session.md
```diff
@@ -8,38 +8,50 @@ In order to configure it, both the framework and core app must set an API key as
 
 - `ApiKeyHeader` - header name that will contain an API key to secure the endpoint added on .NET Framework
 - `ApiKey` - the shared API key that will be validated in the .NET Framework handler
-- `RegisterKey<T>(string)` - Registers a session key to a known type. This is required in order to serialize/deserialize the session state correctly. If a key is found that there is no registration for, an error will be thrown and session will not be available.
```
mjrousos, line 11: Nit-pick, it looks like we removed the explanation of `RegisterKey` here but didn't add it anywhere else as serialization options. It's an important concept, so we should make sure to document it.

docs/session-state/session.md
```diff
@@ -0,0 +1,41 @@
+# Session State
+
+Session state in ASP.NET Framework provided a number of features that ASP.NET Core does not provide. In order to migrate from ASP.NET Framework to Core, the adapters provide mechanisms to enable populating session state with similar behavior as `System.Web` did. Some of the differences between framework and core are:
+
+- ASP.NET Framework would lock session usage so within a session, so subsequent requests in a session are handled in a serial fashion. This is different than ASP.NET Core that does not provide any of these guarantees.
```
mjrousos, line 5:
```suggestion
- ASP.NET Framework would lock session usage within a session, so subsequent requests in a session are handled in a serial fashion. This is different than ASP.NET Core that does not provide any of these guarantees.
```

mjrousos (changes requested):
So, this looks fine for what it is but it's actually kind of the opposite of what I expected design -wise!

This enables users to use the System.Web APIs to access session but have the ASP.NET Core session be the actual backing store. I thought the idea was to enable using ASP.NET Core session APIs but still access data in the remote session.

The reason I see that as useful is because after moving code from the ASP.NET app to the ASP.NET Core one and using existing remote session state functionality to keep things working, users would be able to start gradually changing code that uses session APIs to use the new APIs. Eventually, once everything is moved the shim could be pulled out and the user would just be using ASP.NET Core session.

I can see how this change would also enable that migration, but switching to storing things in ASP.NET Core session means that the user needs to move all of their code to the ASP.NET Core app before they can make use of this.

Maybe it would be useful to sync and talk through the scenario. It's possible I'm just thinking about things backwards!

docs/session-state/remote-session.md
twsouthwick: I thought I had moved it to the serialization section. I've pushed a change with that.

mjrousos (approved):
@twsouthwick and I chatted about this. I think this is useful, it's just a different approach to migrating session than I originally imagined. After talking through the scenario, I think both may be useful in different cases.

This PR enables the scenario of users moving all their code over to ASP.NET Core and then wanting to be able to turn off the remote app without having to remove adapter session usage.

I've opened issue #24 to track the scenario I was imagining. That one involves a user moving a part of their solution to ASP.NET Core and then wanting to update that code to use standard ASP.NET Core APIs (for session access, for example) even if some other code that is still running on ASP.NET needs to share session items with the upgraded code.

I think this PR looks ready to merge and will help with getting an entire code base over onto ASP.NET Core as quickly as possible (the first scenario I described here). We can use #24 to track adding additional features later to support the other scenario.

cc @Tratcher as I'd be interested in hearing any thoughts you have on these topics.

Tratcher:
@mjrousos I agree those are separate scenarios, and both might be useful at different stages. Having an implementation of ISession that exposes the Framework session would help migrate the new controllers off the adapters.

PR #27, opened 2022-05-18
Expand docs and add an overview of the migration journey

twsouthwick:


docs/framework.md
```diff
@@ -0,0 +1,49 @@
...
+}
+```
+
+Customizing the adapters is done by adding modules to via `Application.AddSystemWebAdapters()` which will be loaded within the `SystemWebAdaptersModule` registered in `web.config`. The available configuration right now are described below.
+
+## Additional Modules
+
+### Proxy Support
+
+During migration, the framework app will be moved to be downstream from a reverse proxy. In order for things like port, url, and scheme to be constructed correctly, we must update the values `HttpContext` use. This is similar to the [process used on ASp.NET Core](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balancer).
```
KSchlobohm, line 43: `During migration, the framework app will be moved to be downstream from a reverse proxy.`
could become
`During migration, the framework app will be moved and will be downstream from a reverse proxy.`

docs/intro.md
```diff
@@ -0,0 +1,72 @@
...
+
+```mermaid
+flowchart LR;
+  external[External Traffic] --> core[ASP.NET Core App]
+  core -- Adapters --- libraries
+  core -- YARP proxy --> framework[ \n\nFramework App\n\n\n]
+  framework --- libraries[[Business logic]]
+```
+
+At this point, the journey is to focus on moving routes over one at a time. This could be WebAPI or MVC controllers (or even a single method from a controller), ASPX pages, handlers, or some other implementation of a route. If the route is available in the core app, it will then be matched and served from there. Over time, the core app will start processing more of the routes served than the framework app:
```
KSchlobohm, line 39: At this time, you theoretically have the route in both ASP.NET Core and ASP.NET Framework apps.

I think it would be nice to highlight the advantages of being able to easily test code side-by-side in production (before cutting over). and to be able to easily rollback a release by changing the registered routes.

docs/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+One of the larger challenges is the pervasive use of `System.Web.HttpContext` throughout a code base to access information about a request or update a response. Until now, in order to move forward, a large scale rewrite was required to remove this dependency. Now, the adapters in this repo provide a set of runtime helpers to access the types your code is expecting but in a way that will work on ASP.NET Core with little change.
+
+A complete migration may take a while (sometimes a multi-year effort) depending on the size of the application. In order to continue deploying an application to production while working on migrating, the best pattern is to follow is the [Strangler Fig pattern](https://docs.microsoft.com/en-us/azure/architecture/patterns/strangler-fig). This pattern allows for continual development on the old system with an incremental approach to moving forward. This document will describe how to apply that pattern to an ASP.NET app migrating towards ASP.NET Core.
```
mjrousos, line 7: I wonder if "multi-year effort" is too scary here. For very large applications it may be accurate, but I expect that for many scenarios just saying "months" is still accurate and might not be as likely to scare people off. :) Even with the adapters, it would be nice not to be in a hybrid state for multiple years!

docs/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+```mermaid
+flowchart LR;
+  external[External Traffic] --> core[ASP.NET Core Application]
+  core --- libraries
+  core -- YARP proxy --> framework[ \n\n.NET Framework Application\n\n\n]
+  framework --- libraries[[Business logic]]
+```
+
+In order to start moving over business logic that relies on `HttpContext`, the libraries need to be built against `Microsoft.AspNetCore.SystemWebAdapters`. This will ensure that the libraries are using surface area that is available on both ASP.NET and ASP.NET Core:
```
mjrousos, line 31: I like to include the fact that using SystemWebAdapters can enable libraries to target .NET Standard since that's a known way of making libraries sharable between NetFx and NetCore callers.

docs/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+```mermaid
+flowchart LR;
+  external[External Traffic] --> core[ASP.NET Core Application]
+  core -- Adapters --- libraries
+  core -- YARP proxy --> framework[ \n\n.NET Framework Application\n\n\n]
+  framework --- libraries[[Business logic]]
+```
+
+At this point, the journey is to focus on moving routes over one at a time. This could be WebAPI or MVC controllers (or even a single method from a controller), ASPX pages, handlers, or some other implementation of a route. If the route is available in the core app, it will then be matched and served from there. Over time, the core app will start processing more of the routes served than the .NET Framework Application:
```
mjrousos, line 41: I'm not sure if you want to include it here or not, but as routes are moved it will also be necessary to move services within the app that those endpoints use. For example, the first endpoint that gets moved will probably also involve moving some models, maybe data access stuff, etc.

As that code starts to move, developers will need to start thinking about whether they should (1) duplicate the code, (2) link the code in the new project, or (3) move the code to shared libraries. It might be nice to offer some guidance on those decisions in the docs.

mjrousos (approved):
Looks good! I added a few non-blocking questions/suggestions in inline comments.

README.md
```diff
@@ -2,62 +2,15 @@
...
-}
-```
-
-Notice that since there's a `Controller.Context` property, they can pass that through, but it generally looks the same. Using implicit conversions, the `Microsoft.AspNetCore.Http.HttpContext` can be converted into the adapter that could then be passed around through the levels utilizing the code in the same way.
+To understand how these adapters can be used within the context of a large scale, incremental, migration from ASP.NET to ASP.NET Core, please see the [documentation](docs/README.md).
 
 ## Set up
 Below are the steps needed to start using these adapters in your project:
 
-1. Set up `NuGet.config` to point to the CI feed:
```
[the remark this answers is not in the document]
twsouthwick: Because it will be available on NuGet for the released versions

docs/README.md
twsouthwick: Good point. I'll put those points in, but I'll punt the decision matrix for later

README.md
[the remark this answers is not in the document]
mjrousos: Ah, that makes complete sense. I misread this somehow initially and thought that you meant setting this up was useful for the *users* CI builds but not in other situations and that didn't make sense. Obviously, you mean to use our CI builds they will need NuGet configuration to point at those CI builds. Thanks for the explanation!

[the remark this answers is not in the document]
twsouthwick: Interesting - I wonder if people reading it would have that confusion. If you had that thought, I'm sure others would as well

PR #28, opened 2022-05-18
Simplify remote session module/handlers with tests

twsouthwick:
This change breaks apart the remote session handler into three separate handlers each with a specific function. It also moves any authentication and decisions as to how to handle the request to the module itself. This way, by the time a request has reached the handler, it knows it should just do its thing. Tests have also been included for these as well.


src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/Microsoft.AspNetCore.SystemWebAdapters.SessionState.csproj
```diff
@@ -11,18 +10,13 @@
 
   <ItemGroup>
     <InternalsVisibleTo Include="Microsoft.AspNetCore.SystemWebAdapters.SessionState.Tests" />
+    <InternalsVisibleTo Include="Microsoft.AspNetCore.SystemWebAdapters.SessionState.Framework.Tests" />
     <InternalsVisibleTo Include="DynamicProxyGenAssembly2" Key="0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7" />
```
mjrousos, line 14: I realize this was already here, but do we need this InternalsVisibleTo? It seems like it may not be needed for projects without adapter APIs, so we may as well pull it out.

src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/RemoteSession/Framework/GetWriteableSessionHandler.cs
```diff
@@ -0,0 +1,69 @@
...
+        _serializer = serializer;
+        _cache = cache;
+    }
+
+    public override async Task ProcessRequestAsync(HttpContext context)
+    {
+        using var timeoutCts = new CancellationTokenSource(TimeSpan.FromMinutes(context.Session.Timeout));
+        using var cts = CancellationTokenSource.CreateLinkedTokenSource(timeoutCts.Token, context.Response.ClientDisconnectedToken);
+
+        await ProcessRequestAsync(new HttpContextWrapper(context), cts.Token).ConfigureAwait(false);
```
mjrousos, line 31: Why wrap the context with HttpContextWrapper?

src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/RemoteSession/Framework/ReadOnlySessionHandler.cs
```diff
@@ -0,0 +1,35 @@
...
+    public override bool IsReusable => true;
+
+    public ReadOnlySessionHandler(ISessionSerializer serializer)
+    {
+        _serializer = serializer;
+    }
+
+    public override async Task ProcessRequestAsync(HttpContext context)
+    {
+        await ProcessRequestAsync(new HttpContextWrapper(context));
```
mjrousos, line 24: Same question as before about wrapping the context

src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/Microsoft.AspNetCore.SystemWebAdapters.SessionState.csproj
```diff
@@ -11,18 +10,13 @@
...
   <ItemGroup>
     <FrameworkFiles Include="**/*.Framework.cs" />
-    <StandardFiles Include="**/*.Standard.cs" />
-  </ItemGroup>
-
-  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
-    <Compile Remove="**/*" />
-    <Compile Include="@(StandardFiles)" />
-    <Compile Include="RemoteSession/RemoteAppSessionStateOptions.cs" />
+    <FrameworkFiles Include="RemoteSession/Framework/*.cs" />
```
mjrousos, line 19: Do we want both `*.Framework.cs` and `/Framework/*.cs` denoting netfx-only source? It might be cleaner to just use one or the other.

mjrousos (approved):
Looks good; thanks for cleaning this up!

src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/Microsoft.AspNetCore.SystemWebAdapters.SessionState.csproj
twsouthwick: Yes, this is for testing (all the internals visible are). The DynamicProxyGenAssembly2 is for Castle.Core used by Moq

src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/RemoteSession/Framework/GetWriteableSessionHandler.cs
twsouthwick: Because that is mockable. HttpContext itself cannot be mocked in any way

src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/RemoteSession/Framework/ReadOnlySessionHandler.cs
twsouthwick: Same as above - it's to allow mocking for testing

src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/Microsoft.AspNetCore.SystemWebAdapters.SessionState.csproj
twsouthwick: I have no strong opinion here and am switching to the `Framework/*.cs` as there's enough files.

src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/RemoteSession/Framework/ReadOnlySessionHandler.cs
twsouthwick: Also why the `Application.CompleteRequest()` is not called from the `HttpContextBase` - it's not mockable either.

src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/RemoteSession/Framework/GetWriteableSessionHandler.cs
mjrousos: Ah, interesting. Good to know.

src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/Microsoft.AspNetCore.SystemWebAdapters.SessionState.csproj
mjrousos: Makes sense. Didn't realize this was Moq-related.

Tratcher (approved):

PR #39, opened 2022-05-23
Gracefully handle unknown session keys

### LINKED ISSUE #32: Error reporting when serialization an unknown session key
## Summary

When serializing session, we require keys to be registered with the type they represent. However, people will probably not know the keys they have and will need a way to identify that. The simplest would be to turn on some messaging around keys that aren't registered. This issue is to decide on how we want to do that.

This will focus on the scenario where we have remote app session state enabled.

## Reproduce

Create a remote app session without registering known keys. Attempt to load session on core side. An `InvalidOperationException` is thrown at the point of serializing an unknown key.

- When serializing/deserializing on the framework side, this leads to a 500 error on the framework handler, which propagates down to the core app as a failure. There is no actionable message available here.
- If the framework side has successfully serialized a payload (it knows about all the keys), but the core app encounters a key it is unaware of, it will surface as a `InvalidOperationException` that causes the request to fail. Same thing occurs if a key is added that is unknown when we attempt to serialize. 

### Option 1

This option still causes the request to fail, but attempts to provide the information to mitigate it.

```csharp
public class UnknownSessionKeyException
{
  public IReadOnlyCollection<string> UnknownKeys { get; }
}
```

- The request to the framework app should not result in a 500 error. Instead, it should provide a https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.problemdetails object that lists the unknown keys
- Upon error from the framework app, the core app will raise an UnknownSessionKeyException which lists all the unknown keys
- Upon deserialization/serialization attempt, the session middleware will raise an UnknownSessionKeyException if an unknown key is encountered
- This will still cause a failure of the request on the ASP.NET Core side, but will provide a list of unknown keys

### Option 2

This option will not fail if an unknown key is found, but will log error messages that clearly identify the unknown keys.

- The request to the framework app should not result in a 500 error. If an unknown key is found, it should be added to the payload as a new field that will convey that information to the client.
- When the core app receives its payload, it will right out a message of the unknown keys at an Error level to communicate to the developer that a key was used that has not been registered
- When the session middleware serializes the new keys, it will write out any unknown keys to the log

### Option 3

A combination of Option 1 and Option 3 that can be opted into via the Session options

## Additional thoughts

- Do we want to include the type info of the key that was found if attempting to serialize and it's not registered?

twsouthwick:
This change will cause the serializer to log unknown keys. An option has also been added to throw an exception if unknown keys are found. By default, this will be true, but can is opt-out as needed.

Fixes #32


src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/Serialization/JsonSessionSerializer.Shared.cs
```diff
@@ -2,123 +2,56 @@
...
-            }
-
-            return JsonSerializer.Deserialize(bytes.Span, type, _options);
+            return JsonSerializer.SerializeToUtf8Bytes(value, type, _jsonOptions);
         }
 
-        throw new InvalidOperationException($"Key '{key}' is not registered");
+        LogUnknownKey(key, serialize: true);
+
+        return _options.ThrowOnUnknownSessionKey ? throw new UnknownSessionKeyException(new[] { key }) : Array.Empty<byte>();
```
Tratcher, line 36: Why not always throw?

twsouthwick: I could imagine where someone just wants to log the missing keys so that the system doesn't crash if a new session key is added. However, the flag does allow someone to opt into it.

As an implementation detail, allowing it to be a flag allows the remote app to not throw and pass the info along, while the core app can then decide to throw or not but uses the same implementation.

Tratcher: I would expect the serialization exception to be opt-out. Exceptions are much more discoverable than logs.

twsouthwick: Oh that's reasonable. I've swapped the default

src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/Microsoft.AspNetCore.SystemWebAdapters.SessionState.csproj
```diff
@@ -14,12 +14,11 @@
 
   <ItemGroup>
     <FrameworkFiles Include="**/*.Framework.cs" />
-    <FrameworkFiles Include="RemoteSession/Framework/*.cs" />
+    <FrameworkFiles Include="**/Framework/*.cs" />
```
mjrousos, line 17: Now that we have a Framework folder, can we just move *.Framework.cs there?

src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/Serialization/JsonSessionSerializer.Shared.cs
mjrousos: This looks like a good approach. I can definitely imagine someone wanting to opt out if they have a lot of session items but only expect a few of them to be needed by ASP.NET Core endpoints. I like throwing by default, though.

mjrousos (approved):

src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/Serialization/UnknownSessionKeyException.Shared.cs
```diff
@@ -0,0 +1,27 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.SystemWebAdapters.SessionState.Serialization;
+
+public class UnknownSessionKeyException : Exception
```
Tratcher, line 9:
```suggestion
public class UnknownSessionKeyException : InvalidOperationException
```

Does this new exception type really need to be public? Do we actually expect people to catch it? Or can we use InvalidOperationException instead and put the keys in the message?

We want to avoid introducing new APIs as part of the adapters when people are expected to remove the adapters later.

PR #40, opened 2022-05-23
Enable authenticating with remote app

### LINKED ISSUE #15: Auth: Cookie auth

mjrousos:
This PR enables remote authentication so that the ASP.NET Core app in an upgrade workflow can defer to the original ASP.NET app for user identity.

Fixes #15


src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationAuthHandler.cs
```diff
@@ -0,0 +1,107 @@
...
+/// <summary>
+/// Authentication handler that authenticates users by making requests to a remote app
+/// for authentication via a remote authentication service.
+/// </summary>
+internal class RemoteAuthenticationAuthHandler : AuthenticationHandler<RemoteAuthenticationOptions>
+{
+    private readonly IRemoteAuthenticationService _authService;
+    private readonly IEnumerable<IRemoteAuthenticationResultProcessor> _resultProcessors;
+    private readonly ILogger<RemoteAuthenticationAuthHandler> _logger;
+    private RemoteAuthenticationResult? _remoteAuthenticationResult;
```
twsouthwick, line 25: Are handlers a transient service? I thought they'd be singleton

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationAuthHandler.cs
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        return _remoteAuthenticationResult;
+    }
+
+    protected override async Task<AuthenticateResult> HandleAuthenticateAsync()
+    {
+        // If remote authentication is enabled for the endpoint, retrieve the remote authentication results and return
+        // an appropriate AuthenticateResult value.
+        if (Context.GetEndpoint()?.Metadata.GetMetadata<IRemoteAuthenticationMetadata>() is IRemoteAuthenticationMetadata metadata && metadata.Enabled)
```
twsouthwick, line 65:
```suggestion
        if (Context.GetEndpoint()?.Metadata.GetMetadata<IRemoteAuthenticationMetadata>() is { Enabled: true})
```

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationExtensions.cs
```diff
@@ -0,0 +1,81 @@
...
+/// Helper methods for registering remote authentication services
+/// </summary>
+public static class RemoteAuthenticationExtensions
+{
+    /// <summary>
+    /// Registers remote authentication auth handler with an authentication builder using a default scheme name.
+    /// </summary>
+    /// <param name="authenticationBuilder">The authentication builder to register the remote authentication handler with.</param>
+    /// <returns>The authentication builder updated with the remote authentication handler added.</returns>
+    public static AuthenticationBuilder AddRemote(this AuthenticationBuilder authenticationBuilder)
```
twsouthwick, line 22: We've been referring to it as `RemoteApp` for session - does it make sense to stay consistent there?

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationHttpHandler.Framework.cs
```diff
@@ -0,0 +1,62 @@
...
+        //        about the currently authenticated user.
+        if (!string.Equals(_options.RemoteServiceOptions.ApiKey, context.Request.Headers.Get(_options.RemoteServiceOptions.ApiKeyHeader), StringComparison.Ordinal))
+        {
+            // Using 407 here (proxy authentication required) to differentiate from the scenario of
+            // a valid API key but no authenticated user.
+            context.Response.StatusCode = 407;
+        }
+        else
+        {
+            // If a user is logged in (using ASP.NET's usual authenticaiton mechanisms), return that claims principal.
```
twsouthwick, line 47:
```suggestion
            // If a user is logged in (i.e. using ASP.NET's usual authentication mechanisms), return that claims principal.
```

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationHttpHandler.Framework.cs
```diff
@@ -0,0 +1,62 @@
...
+            if (context.User is ClaimsPrincipal claimsPrincipal && context.User.Identity.IsAuthenticated)
+            {
+                context.Response.StatusCode = 200;
+                context.Response.ContentType = "application/octet-stream";
+                claimsPrincipal.WriteTo(new BinaryWriter(context.Response.OutputStream));
+            }
+            else
```
twsouthwick, line 54: should we return a different error code if there is a `context.User` but it is not a `ClaimsPrincipal`?

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationOptions.cs
```diff
@@ -0,0 +1,73 @@
...
+    : AuthenticationSchemeOptions
+#endif
+{
+#if NETCOREAPP3_1_OR_GREATER
+    public static readonly string[] DefaultHeadersToForward = new[]
+    {
+        "Authorization"
+    };
+
+    public static readonly string[] DefaultResponseHeadersToForward = new[]
```
twsouthwick, line 25: Do we need to make them public? Array values could be modified here - if we need it public, can you expose it as a `IReadOnlyCollection`?

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationResultFactory.cs
```diff
@@ -0,0 +1,65 @@
...
+    public async Task<RemoteAuthenticationResult> CreateRemoteAuthenticationResultAsync(HttpResponseMessage response, RemoteAuthenticationOptions options)
+    {
+        RemoteAuthenticationResult? ret = null;
+
+        // If the result has a 200 status code, attempt to deserialize the ClaimsPrincipal
+        if (response.StatusCode == HttpStatusCode.OK)
+        {
+            try
+            {
+                using var reader = new BinaryReader(await response.Content.ReadAsStreamAsync());
```
twsouthwick, line 37: the stream should be disposed

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationResultFactory.cs
```diff
@@ -0,0 +1,65 @@
...
+                ret = new RemoteAuthenticationResult(new ClaimsPrincipal(reader), (int)response.StatusCode);
+            }
+            catch (Exception exc)
+            {
+                _logger.LogWarning(exc, "Failed to deserialize remote authentication response as a claims principal despite success status code");
```
twsouthwick, line 42: Do we want to silently ignore? Or do we want to log and throw?

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationResultFactory.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        // Copy expected response headers
+        var headersToForward = response.Headers.Select(h => h.Key);
+        if (options.ResponseHeadersToForward.Any())
+        {
+            headersToForward = headersToForward.Where(headersToForward.Contains);
+        }
+        foreach (var headerName in headersToForward)
+        {
+            ret.ResponseHeaders.Add(headerName, response.Headers.GetValues(headerName));
```
twsouthwick, line 60: This seems to iterate a few times over the headers via LINQ - can you just do a single pass?

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationService.cs
```diff
@@ -0,0 +1,131 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.AspNetCore.SystemWebAdapters.Authentication;
+
+/// <summary>
+/// Service for authenticating a user by making an HTTP request on their behalf to a remote app.
+/// </summary>
+public class RemoteAuthenticationService : IRemoteAuthenticationService
```
twsouthwick, line 21: services have for now been kept internal with only the interfaces (as needed) being exposed. Is there a reason to expose this service itself?

twsouthwick (commented):
Awesome! Can you add some documentation about what the flow here looks like? (kind of like the remote session stuff).

mjrousos:
Yep, for sure. I'll get some docs added today.

twsouthwick:
Question: Do we want this in a separate library? The session stuff was put in a separate package because it would force new dependencies brought in by System.Text.Json. Does this bring in new dependencies?

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/IRemoteAuthenticationService.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// <summary>
+    /// A service for authenticating an HTTP request with a remote service.
+    /// </summary>
+    public interface IRemoteAuthenticationService
+    {
+        /// <summary>
+        /// Initializes the remote authentication service for the given scheme.
+        /// </summary>
+        /// <param name="scheme">The scheme whose configuration the remote authentication service should use for authenticating requests.</param>
+        void Initialize(AuthenticationScheme scheme);
```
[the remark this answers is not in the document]
mjrousos: That would work. I chose this approach because it mirrors what `AuthenticationHandler` (which depends on the service) does. Having the initialize here makes it straightforward to have `RemoteAuthenticationAuthHandler`'s initialize method call the service's initialize method.

An alternative would be to have the `RemoteAuthenticationAuthHandler` have a dependency on a service factory, instead, and create the service in its Initialize method.

I don't have a strong preference one way or the other. If you think the factory approach would be cleaner/more maintainable/whatever, let me know and I'll change it.

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/NullableAttributes.cs
```diff
@@ -0,0 +1,180 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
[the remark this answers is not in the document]
mjrousos: Good point. It's already in the core web adapters project. I'll link to it there.

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationHttpHandler.Framework.cs
mjrousos: That's a good question. I like that idea but I'm not sure what status code would make sense, though. Maybe 500, but that's pretty general. Maybe 409 or 422 but neither of those seems quite right.

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationModule.Framework.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        var handler = new RemoteAuthenticationHttpHandler(_options);
+
+        context.PostMapRequestHandler += MapRemoteAuthenticationHandler;
+
+        void MapRemoteAuthenticationHandler(object sender, EventArgs e)
+        {
+            var context = ((HttpApplication)sender).Context;
+
+            if (string.Equals(context.Request.Path, _options.AuthenticationEndpointPath)
```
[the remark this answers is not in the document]
mjrousos: Sure. I'll move them here.

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationOptions.cs
mjrousos: Good point. I suppose these don't need to be public, but it seems nice-to-have to have them available (maybe users will want to take existing default response headers to forward and add a couple to make their own list). I'll make these ImmutableArrays.

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationResult.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// Gets the status code returned in the remote authentication result.
+    /// If a user was successfully retrieved, this status code will be 200.
+    /// </summary>
+    public int StatusCode { get; }
+
+    /// <summary>
+    /// Gets a dictionary of auth-related headers that may need propagated back
+    /// to the caller if remote authentication fails.
+    /// </summary>
+    public IDictionary<string, IEnumerable<string>> ResponseHeaders { get; } = new Dictionary<string, IEnumerable<string>>();
```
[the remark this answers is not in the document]
mjrousos: Good call; switching.

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationResultFactory.cs
mjrousos: Fixed

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationResultFactory.cs
mjrousos: I think we don't need to throw here. Even if there's no ClaimsPrincipal, we can still report the status code and headers. Likely something's wrong on the server side, but I lean towards allowing the client to still work (just without remote auth) and providing log messages.

If it turns out that this comes up a lot and we need to make the failure more obvious, we could switch to throwing.

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationResultFactory.cs
mjrousos: I simplified this (and fixed a bug)

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationService.cs
```diff
@@ -0,0 +1,131 @@
...
+    {
+        if (!Initialized)
+        {
+            _logger.LogError("Remote authentication handler must be initialized before authenticating");
+            throw new InvalidOperationException("Remote authentication handler must be initialized before authenticating");
+        }
+
+        // Create a new HTTP request, but propagate along configured headers or cookies
+        // that may matter for authentication
+        var authRequest = new HttpRequestMessage();
```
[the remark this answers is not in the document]
mjrousos: Good catch. Fixed.

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RedirectUrlProcessor.cs
```diff
@@ -0,0 +1,68 @@
...
+                    var queryStrings = QueryHelpers.ParseQuery(redirectLocation.Query);
+                    if (queryStrings.ContainsKey(ReturnUrlQueryStringName))
+                    {
+                        // Get the plain redirect URL without the query string
+                        var redirectWithoutQuery = redirectLocation.ToString().Replace(redirectLocation.Query, string.Empty);
+
+                        // Update the query strings to use the original request path as the ReturnUrl query string
+                        queryStrings[ReturnUrlQueryStringName] = context.Request.Path.Value;
+
+                        // .NET Core 3.1 doesn't have a QueryHelpers.AddQueryString that takes an argument of type Dictionary<string, StringValues>
```
Tratcher, line 52: Aren't we planning to remove 3.1 for reasons like this?

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RedirectUrlProcessor.cs
```diff
@@ -0,0 +1,68 @@
...
+    /// Updates response headers in a remote authentication result so that any
+    /// 'Location' headers will include a ReturnUrl query string corresponding to the
+    /// URL of the original request rather than the URL of the remote authentication endpoint.
+    /// </summary>
+    /// <param name="result"></param>
+    /// <param name="context"></param>
+    /// <returns></returns>
+    public Task ProcessAsync(RemoteAuthenticationResult result, HttpContext context)
+    {
+        if (result.ResponseHeaders.TryGetValue(LocationHeaderName, out var locationHeaders))
```
Tratcher, line 34: Set-Cookie headers can have a similar problem with their path and domain fields.

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationAuthHandler.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            var authResult = await GetRemoteAuthenticationResultAsync();
+
+            if (authResult.User is not null)
+            {
+                var ticket = new AuthenticationTicket(authResult.User, Scheme.Name);
+                _logger.LogDebug("Authenticated user {UserName} based on remote authentication service", authResult.User.Identity?.Name);
```
Tratcher, line 72: PII, remove the username.

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationAuthHandler.cs
```diff
@@ -0,0 +1,107 @@
...
+            var authResult = await GetRemoteAuthenticationResultAsync();
+
+            // Propagate headers and status code back to the caller
+            // Different authentication schemes may challenge in different ways in the remote
+            // app, so make a best effort to forward the effects of these challenges by forwarding
+            // configured headers (like Location, perhaps) and status code (like 302 or 401, for example).
+            Context.Response.StatusCode = authResult.StatusCode;
+            foreach (var header in authResult.ResponseHeaders)
+            {
+                Context.Response.Headers.Add(header);
```
Tratcher, line 103: IHeaderDictionary.Add is a problematic API we're discourage using. E.g. it throws if you have multiple headers with the same name.
```suggestion
                Context.Response.Headers.Append(header.Key, header.Value);
```

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationExtensions.cs
```diff
@@ -0,0 +1,81 @@
...
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.AspNetCore.SystemWebAdapters.Authentication;
+
+/// <summary>
+/// Helper methods for registering remote authentication services
+/// </summary>
+public static class RemoteAuthenticationExtensions
```
Tratcher, line 15: RemoteAuthentication* -> RemoteAppAuthentication* to avoid conflicts with existing types.

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationModule.Framework.cs
```diff
@@ -0,0 +1,51 @@
...
+                && context.Request.HttpMethod.Equals("GET", StringComparison.OrdinalIgnoreCase))
+            {
+                // TODO : Do we need to even check the API key for this? The handler only returns information
+                //        about the currently authenticated user.
```
Tratcher, line 32: Yes, otherwise you risk leaking PII

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationModule.Framework.cs
```diff
@@ -0,0 +1,51 @@
...
+                if (!string.Equals(_options.RemoteServiceOptions.ApiKey, context.Request.Headers.Get(_options.RemoteServiceOptions.ApiKeyHeader), StringComparison.Ordinal))
+                {
+                    // Using 407 here (proxy authentication required) to differentiate from the scenario of
+                    // a valid API key but no authenticated user.
+                    context.Response.StatusCode = 407;
```
Tratcher, line 37: It looks like this status code would be blindly returned by the front end to the client. We don't want to send them 407s.
400 bad request or 500 internal server error.

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationOptions.cs
```diff
@@ -0,0 +1,74 @@
...
+
+namespace Microsoft.AspNetCore.SystemWebAdapters.Authentication;
+
+public class RemoteAuthenticationOptions
+#if NETCOREAPP3_1_OR_GREATER
+    : AuthenticationSchemeOptions
+#endif
+{
+#if NETCOREAPP3_1_OR_GREATER
+    public static readonly IEnumerable<string> DefaultHeadersToForward = ImmutableArray.CreateRange(new[]
```
Tratcher, line 21:
```suggestion
    public static readonly IEnumerable<string> DefaultRequestHeadersToForward = ImmutableArray.CreateRange(new[]
```

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationOptions.cs
```diff
@@ -0,0 +1,74 @@
...
+    /// Gets or sets the remote service options used to connect with the remote authentication service.
+    /// </summary>
+    public RemoteServiceOptions RemoteServiceOptions { get; set; } = new RemoteServiceOptions();
+
+#if NETCOREAPP3_1_OR_GREATER
+    /// <summary>
+    /// Gets or sets a list of request headers that should be forwarded to the remote app for authentication purposes. If no headers
+    /// are specified, all headers will be forwarded.
+    /// </summary>
+    public IList<string> HeadersToForward { get; set; } = new List<string>(DefaultHeadersToForward);
```
Tratcher, line 44:
```suggestion
    public IList<string> RequestHeadersToForward { get; set; } = new List<string>(DefaultHeadersToForward);
```

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationService.cs
```diff
@@ -0,0 +1,131 @@
...
+        foreach (var headerName in headerNames)
+        {
+            authRequest.Headers.Add(headerName, originalRequest.Headers[headerName].ToArray());
+        }
+    }
+
+    // Add configured cookies to the request, or all cookies if none in particualr are specified
+    private static void AddCookies(IEnumerable<string> cookiesToForward, HttpRequest originalRequest, HttpRequestMessage authRequest)
+    {
+        IEnumerable<string> cookieNames = originalRequest.Cookies.Keys;
```
Tratcher, line 114: Why filter and re-build the cookies? It would be simpler and more efficient to forward the whole Cookie header using the header forwarding logic.

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationHttpHandler.Framework.cs
```diff
@@ -0,0 +1,51 @@
...
+        // If a user is logged in (using ASP.NET's usual authenticaiton mechanisms), return that claims principal.
+        if (context.User is ClaimsPrincipal claimsPrincipal && context.User.Identity.IsAuthenticated)
+        {
+            context.Response.StatusCode = 200;
+            context.Response.ContentType = "application/octet-stream";
+            claimsPrincipal.WriteTo(new BinaryWriter(context.Response.OutputStream));
+        }
+        else
+        {
+            context.Response.StatusCode = 401;
```
Tratcher, line 46: How do you expect the module to set challenges like 301 with a location header? Or 401 with a www-authenticate header?

Tratcher (changes requested):
Which kinds of auth have you tried this with?

mjrousos:
@twsouthwick - that's a good point. This doesn't pull in any new dependencies, so we could include it in the SystemWebAdapters project without adding a new one. I suppose it might be nice for customers to not have too many packages they need to install.

@Tratcher - so far, I've tested it with OWIN cookie auth (ASP.NET Idenity) and it works for that scenario. I'll try out forms auth and JWT tokens to confirm it works for those scenarios, too.

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/IRemoteAuthenticationMetadata.cs
```diff
@@ -0,0 +1,17 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.AspNetCore.SystemWebAdapters.Authentication;
+
+/// <summary>
+/// Metadata for determining whether the remote authentication handler should
+/// be enabled for an endpoint or not. Remote authentication enables useful migration
+/// scenarios but adds unnecessary performance overhead in cases where it isn't needed.
+/// </summary>
+public interface IRemoteAuthenticationMetadata
```
davidfowl, line 11: Why do we need this? Can't we treat it like any other auth handler?

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RedirectUrlProcessor.cs
twsouthwick: It's removed now - you can use all the .NET 6 goodness you want :)

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationService.cs
mjrousos: I was thinking that some users may want to forward only specific cookies (auth-related ones). I'm fine pulling this out, though, and just treating the cookie header as another header to forward. We can always revisit if we hear feedback from customers that they need more control.

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/IRemoteAuthenticationMetadata.cs
davidfowl: Can we remove this?

mjrousos: Yea, I'll pull this out and we can just have options to either make this scheme the default scheme (which would be like enabling the metadata everywhere) or not (in which case, users will have to specify specific endpoints this scheme is used with).

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationModule.Framework.cs
mjrousos: 500 may be the best option since 400 implies something wrong with the user's request when really it's an issue with proxy setup. I'll use that.

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationOptions.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.ComponentModel.DataAnnotations;
+using System.Collections.Immutable;
+#endif
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.SystemWebAdapters.Authentication;
+
+public class RemoteAuthenticationOptions
```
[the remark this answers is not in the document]
mjrousos: I'll use RemoteAppAuthenticationOptions instead.

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationHttpHandler.Framework.cs
mjrousos: At least for the OWIN scenarios I've tried so far, the OWIN middleware for cookie auth or bearer auth actually takes care of that. Those components just look for a 401 response (which I set here) and update headers, status code, etc. as needed. I still need to confirm that this will also "magically work" for forms auth scenarios but will look into that this week.

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationHttpHandler.Framework.cs
mjrousos: I like that approach I'll do that.

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationAuthHandler.cs
mjrousos: Fixed

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationAuthHandler.cs
mjrousos: Fixed

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationModule.Framework.cs
mjrousos: Thinking more about this, we don't want any of these responses to get back to the user. I'll leave this as 407 but update the client (`RemoteAppAuthenticationAuthHandler`) to understand 407 and log the issue but not propagate the status code back to the user.

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationService.cs
mjrousos: I've removed this and added "Cookie" to the list of request headers forwarded by default. If we find users need the ability to forward some cookies but not others, we can revisit this.

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/IRemoteAuthenticationMetadata.cs
mjrousos: Ok, this is removed now. Instead, users specify whether to make remote auth the default or not when registering SystemWebAdapter services for it.

docs/remote-authentication/remote-authentication.md
```diff
@@ -0,0 +1,81 @@
...
+    .AddRemoteAppAuthentication(options =>
+    {
+        options.RemoteServiceOptions.RemoteAppUrl = new("http://URL-for-the-ASPNet-app");
+        options.RemoteServiceOptions.ApiKey = "MySecretKey";
+    });
+```
+
+In addition to configuring the remote app's URL and the shared secret ApiKey, the callback passed to `AddRemoteAppAuthentication` can also optionally specify some aspects of the remote authentication process's behavior:
+
+* `HeadersToForward`: This property contains headers that should be forwarded from a request when calling the authenticate API. By default, the only header forwarded is `Authorization`. If no headers are specified (including removing the default one), then all headers will be forwarded.
```
Tratcher, line 38:
```suggestion
* `RequestHeadersToForward`: This property contains headers that should be forwarded from a request when calling the authenticate API. By default, the only header forwarded is `Authorization`. If no headers are specified (including removing the default one), then all headers will be forwarded.
```

docs/remote-authentication/remote-authentication.md
```diff
@@ -0,0 +1,81 @@
...
+* `CookiesToForward`: This property contains cookies that should be forwarded from a request when calling the authenticate API. By default, all cookies are forwarded, but adding cookie names to this list will cause only those cookies to be used.
```
Tratcher, line 39:
```suggestion
```

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationExtensions.cs
mjrousos: Done

mjrousos:
@twsouthwick @Tratcher @davidfowl - Thanks for the reviews on this one. I think at this point I've addressed all the feedback so far. I've combined the auth code into the main System.Web adapters project and removed the remote auth metadata (depending, instead, on existing default scheme settings).

Can you please take another look and let me know what other feedback you have?

The only item still outstanding is that I need to test this with forms auth (I've tested it with cookies and bearer tokens). I'll work on that this week and get an additional sample added showing forms auth working.

docs/remote-authentication/remote-authentication.md
```diff
@@ -0,0 +1,80 @@
...
+        options.RemoteServiceOptions.ApiKey = "MySecretKey";
+    });
+```
+
+In addition to configuring the remote app's URL and the shared secret ApiKey, the callback passed to `AddRemoteAppAuthentication` can also optionally specify some aspects of the remote authentication process's behavior:
+
+* `RequestHeadersToForward`: This property contains headers that should be forwarded from a request when calling the authenticate API. By default, the only header forwarded is `Authorization`. If no headers are specified (including removing the default one), then all headers will be forwarded.
+* `ResponseHeadersToForward`: This property lists response headers that should be propagated back from the Authenticate request to the original call that prompted authentication in scenarios where identity is challenged. By default, this includes `Location`, `Set-Cookie`, and `WWW-Authenticate` headers.
+* `AuthenticationEndpointPath`: The endpoint on the ASP.NET app where authenticate requests should be made. This defaults to `/systemweb-adapters/authenticate` and must match the endpoint specified in ASP.NET authentication endpoint configuration.
```
twsouthwick, line 40:
```suggestion
* `AuthenticationEndpointPath`: The endpoint on the ASP.NET app where authenticate requests should be made. This defaults to `/systemweb-adapters/authenticate` and must match the endpoint specified in the  ASP.NET authentication endpoint configuration.
```

samples/RemoteAuth/Bearer/RemoteBearer/App_Start/OpenIdConnectCachingSecurityTokenProvider.cs
```diff
@@ -0,0 +1,93 @@
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.IdentityModel.Protocols;
+using Microsoft.IdentityModel.Protocols.OpenIdConnect;
+using Microsoft.IdentityModel.Tokens;
+using Microsoft.Owin.Security.Jwt;
+
+// Coped from https://github.com/Azure-Samples/active-directory-b2c-dotnet-webapp-and-webapi
```
twsouthwick, line 9:
```suggestion
// Copied from https://github.com/Azure-Samples/active-directory-b2c-dotnet-webapp-and-webapi
```

samples/RemoteAuth/Bearer/RemoteBearer/App_Start/OpenIdConnectCachingSecurityTokenProvider.cs
```diff
@@ -0,0 +1,93 @@
...
+
+namespace RemoteOAuth
+{
+    // This class is necessary because the OAuthBearer Middleware does not leverage
+    // the OpenID Connect metadata endpoint exposed by the STS by default.
```
twsouthwick, line 14: what's STS? should that be spelled out?

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/ForwardedHost.Shared.cs
```diff
@@ -0,0 +1,29 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+
+namespace Microsoft.AspNetCore.SystemWebAdapters;
+
+internal struct ForwardedHost
```
twsouthwick, line 8:
```suggestion
internal readonly struct ForwardedHost
```

src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/IAuthenticationResultFactory.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.SystemWebAdapters.Authentication;
+
+/// <summary>
+/// Interface for creating authentication results from HTTP responses returned
+/// by a remote app in remote authentication scenarios.
+/// </summary>
+public interface IAuthenticationResultFactory
```
twsouthwick, line 13: Does this need to be public? Or is it an implementation detail? Do we expect people to need to override this?

src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/IRemoteAppAuthenticationResultProcessor.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Microsoft.AspNetCore.SystemWebAdapters.Authentication;
+
+/// <summary>
+/// An interface for types that can modify a remote authentication result after
+/// it has been retrieved from a remote service. This interface allows users
+/// to specify heuristics to apply to remote authentication results to clean them up
+/// or modify their properties in an extensible way.
+/// </summary>
+public interface IRemoteAppAuthenticationResultProcessor
```
twsouthwick, line 15: Do we expect people to need to implement this?

src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/RemoteAppAuthenticationHttpHandler.Framework.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.IO;
+using System.Security.Claims;
+using System.Web;
+
+namespace Microsoft.AspNetCore.SystemWebAdapters.Authentication;
+
+/// <summary>
+/// HTTP handler for serving requests to remote authenticaiton endpoint.
+/// </summary>
+internal sealed class RemoteAppAuthenticationHttpHandler : IHttpHandler
```
twsouthwick, line 14: we could implement `IHttpHandler` and `IHttpModule` together for this to keep the code together. If not, could the API key check be moved to the module? That way it fails early

twsouthwick (commented):
This looks great - had a few questions about visibility of some of the classes and some docs changes.

It would be great to see some tests for some of the processors to understand what it's expected to do. Not a blocker, though

mjrousos:
I've added a forms auth sample demonstrating that remote auth (more or less) works with forms auth, too. Like cookie auth, logging out from the ASP.NET Core app doesn't work (because forms auth logout does some magic that I haven't implemented on the ASP.NET Core side yet) but everything else works.

Tratcher:
> Like cookie auth, logging out from the ASP.NET Core app doesn't work (because forms auth logout does some magic that I haven't implemented on the ASP.NET Core side yet) but everything else works.

What kind of magic?

mjrousos:
The LoginStatus element of the forms app (`<asp:LoginStatus ID="LoginStatus1" runat="server" LogoutAction="Redirect" LogoutPageUrl="~/Logout.aspx" />`) generates a logout link that logs the user out via a post back (`__doPostBack('ctl00$LoginStatus1$ctl00','')`).

I expect the user could mimic this action from the ASP.NET Core app, but my guess is they'll need to make a request that looks like a post back and I haven't dug into what all that entails.

mjrousos:
Thanks, @Tratcher. Those are good questions. I agree that this isn't going to be the right approach for all auth scenarios (JWT can be done better pretty easily and we can potentially provide some code to help with cookies). I think this is a useful first step to getting auth working that can be used as a fallback for most scenarios (everything except Windows auth, I think) until we have a better solution or better guidance available.

As far as how to migrate off of it, that's a good point. I think we'll probably want a follow-up feature that allows users to authenticate in the *ASP.NET Core* app and forward the identity to the ASP.NET app (so, the reverse of what's happening here).  I've already prototyped that approach and it works. That can be a separate PR, though, and I think this one is the higher priority based on a principle of getting people started with .NET 6 as quickly and painlessly as possible.

samples/RemoteAuth/Bearer/RemoteBearer/packages.config
```diff
@@ -0,0 +1,37 @@
...
+  <package id="Microsoft.AspNet.WebApi.HelpPage" version="5.2.7" targetFramework="net472" />
+  <package id="Microsoft.AspNet.WebApi.WebHost" version="5.2.7" targetFramework="net472" />
+  <package id="Microsoft.AspNet.WebPages" version="3.2.7" targetFramework="net472" />
+  <package id="Microsoft.CodeDom.Providers.DotNetCompilerPlatform" version="2.0.1" targetFramework="net472" />
+  <package id="Microsoft.IdentityModel.Abstractions" version="6.18.0" targetFramework="net472" />
+  <package id="Microsoft.IdentityModel.JsonWebTokens" version="6.18.0" targetFramework="net472" />
+  <package id="Microsoft.IdentityModel.Logging" version="6.18.0" targetFramework="net472" />
+  <package id="Microsoft.IdentityModel.Protocols" version="6.18.0" targetFramework="net472" />
+  <package id="Microsoft.IdentityModel.Protocols.OpenIdConnect" version="6.18.0" targetFramework="net472" />
+  <package id="Microsoft.IdentityModel.Protocols.WsFederation" version="5.3.0" targetFramework="net472" />
```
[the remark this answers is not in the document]
mjrousos: Huh, not sure. I'll update them all to 6.19.

src/Microsoft.AspNetCore.SystemWebAdapters.Authentication/RemoteAuthenticationHttpHandler.Framework.cs
mjrousos: I've confirmed that this works for forms auth, too, so at least for most common kinds of auth the 401 returned by the handler is translated into the appropriate status code and headers.

samples/RemoteAuth/Bearer/RemoteBearer/App_Start/OpenIdConnectCachingSecurityTokenProvider.cs
```diff
@@ -0,0 +1,93 @@
...
+                    return _keys;
+                }
+                finally
+                {
+                    _synclock.ExitReadLock();
+                }
+            }
+        }
+
+        private void RetrieveMetadata()
```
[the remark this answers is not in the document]
mjrousos: Fixed

src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/ResultProcessors/RedirectUrlProcessor.cs
```diff
@@ -0,0 +1,116 @@
...
+/// <summary>
+/// Processes remote authentication results to fix-up 'ReturnUrl' parameters
+/// given in redirect responses so that they will return the user to the correct
+/// original URL instead of to the remote authentication endpoint.
+/// </summary>
+internal class RedirectUrlProcessor : IRemoteAppAuthenticationResultProcessor
+{
+    private const string LocationHeaderName = "Location";
+    private const string ReturnUrlQueryStringName = "ReturnUrl";
+    private const string ForwardedHostHeaderName = "x-forwarded-host";
```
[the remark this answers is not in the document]
mjrousos: Not in `HeaderNames`, unfortunately, but I like the idea of using a shared const for this. I'll do that.

src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/RemoteAppAuthenticationService.cs
```diff
@@ -0,0 +1,120 @@
...
+        LogAuthenticationResponse(response.StatusCode);
+
+        return await _resultFactory.CreateRemoteAppAuthenticationResultAsync(response, _options);
+    }
+
+    // Add configured headers to the request, or all headers if none in particular are specified
+    private static void AddHeaders(IEnumerable<string> headersToForward, HttpRequest originalRequest, HttpRequestMessage authRequest)
+    {
+        // Add x-forwarded headers so that the authenticate API will
+        authRequest.Headers.Add("x-forwarded-host", originalRequest.Host.Value);
```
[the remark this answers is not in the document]
mjrousos: No, unfortunately.

src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/RemoteAppAuthenticationOptions.Shared.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Collections.Immutable;
+#endif
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.SystemWebAdapters.Authentication;
+
+public class RemoteAppAuthenticationOptions
+#if NET6_0_OR_GREATER
```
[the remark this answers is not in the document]
mjrousos: That doesn't work here because we need to exclude .NET Standard, too.

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/ForwardedHost.Shared.cs
mjrousos: Fixed.

samples/RemoteAuth/Bearer/RemoteBearer/App_Start/OpenIdConnectCachingSecurityTokenProvider.cs
mjrousos: Fixed

docs/getting_started.md
```diff
@@ -39,6 +39,10 @@ This step may require a number of projects to change depending on your solution
 
 Session is a commonly used feature of ASP.NET that shares a name with a feature in ASP.NET Core but that's where the similarity ends. Please see the documentation on [session support](session-state/session.md) to understand how to use it.
 
+## Enable shared authentication support
+
+While migrating endpoints from an ASP.NET app to an ASP.NET Core app, it is desirable to enable the user to log in to the application once (if authentication is supported) and for that identity to be available to both the ASP.NET and ASP.NET Core app so that the user's experience with the site works as it would if all endpoints were served from the same project. In order to enable this, the System.Web adapters expose [remote authentication features](remote-authentication/remote-authentication.md) for deferring authentication to the ASP.NET app and sharing the user's ClaimsPrincipal between both apps.
```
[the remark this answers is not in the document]
mjrousos: I think it's important we (briefly) discuss shared auth in the getting started and link to the page with more details. I'll see if I can simplify this.

docs/remote-authentication/remote-authentication.md
mjrousos: Fixed

docs/remote-authentication/remote-authentication.md
```diff
@@ -0,0 +1,80 @@
...
+
+In addition to these options, `RemoteAuthenticationOptions` derives from `AuthenticationSchemeOptions`, so it can be used to optionally specify authentication schemes to forward to for different authentication actions.
+
+In addition to configuring services for the remote authentication handler, **endpoints must opt in to using remote authentication**. This can be done for an entire set of endpoints at once by calling `RequireRemoteAppAuthentication` while configuring the endpoints:
```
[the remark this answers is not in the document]
mjrousos: Good catch. I forgot to update this part of the docs when I made that change. I'll revise them.

docs/remote-authentication/remote-authentication.md
```diff
@@ -0,0 +1,80 @@
...
+    1. If remote authentication is enabled for the endpoint, the handler will make an HTTP request to the ASP.NET app's authenticate endpoint. It will copy configured headers and cookies from the current request onto this new one in order to forward auth-relevant data. As mentioned above, default behavior is to copy the `Authorize` header and all cookies. The API key header is also added for security purposes.
+1. The ASP.NET app's `RemoteAuthenticationHttpHandler` will serve requests sent to the authenticate endpoint. As long as the API keys match, the handler will return either the current user's ClaimsPrincipal serialized into the response body (the use should have already been logged in thanks to the forwarded headers and cookies) or it will return 401. ASP.NET authentication middleware will intercept the 401 response and either add relevant response headers (`WWW-Authenticate`, for example) or change the response to a 302 with a `Location` header to indicate where the requester should redirect to.
```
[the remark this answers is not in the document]
mjrousos: Removed

src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/IAuthenticationResultFactory.cs
mjrousos: I think it's unlikely users will override this, so I've made it internal.

src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/IRemoteAppAuthenticationResultProcessor.cs
mjrousos: This one, I think users may want to implement. If their ASP.NET app has any sort of custom authentication that would return different headers or status codes than expected, implementing this interface allows the user to tweak the response object accordingly.

src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/IRemoteAppAuthenticationService.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.SystemWebAdapters.Authentication
+{
+    /// <summary>
+    /// A service for authenticating an HTTP request with a remote service.
+    /// </summary>
+    public interface IRemoteAppAuthenticationService
```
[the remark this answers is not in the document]
mjrousos: No, good point. I'll make it internal.

src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/RemoteAppAuthenticationAuthHandler.cs
```diff
@@ -0,0 +1,104 @@
...
+        _logger = loggerFactory.CreateLogger<RemoteAppAuthenticationAuthHandler>();
+        _authService = authService;
+        _resultProcessors = resultProcessors;
+    }
+
+    protected override Task InitializeHandlerAsync() => _authService.InitializeAsync(Scheme);
+
+    private async Task<RemoteAppAuthenticationResult> GetRemoteAppAuthenticationResultAsync()
+    {
+        if (_remoteAppAuthResult is null)
```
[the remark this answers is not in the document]
mjrousos: We cache the result because we need the remote app authentication result both for `HandleAuthenticateAsync` and for `HandleChallengeAsync`, so this may be called twice.

src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/RemoteAppAuthenticationExtensions.cs
```diff
@@ -0,0 +1,92 @@
...
+        if (authenticationBuilder is null)
+        {
+            throw new ArgumentNullException(nameof(authenticationBuilder));
+        }
+
+        authenticationBuilder.Services.AddScoped<IRemoteAppAuthenticationResultProcessor, RedirectUrlProcessor>();
+        authenticationBuilder.Services.AddSingleton<IAuthenticationResultFactory, RemoteAppAuthenticationResultFactory>();
+        authenticationBuilder.Services.AddHttpClient<IRemoteAppAuthenticationService, RemoteAppAuthenticationService>()
+            // Disable cookies in the HTTP client because the service will manage the cookie header directly
+            .ConfigurePrimaryHttpMessageHandler(() => new HttpClientHandler { UseCookies = false, AllowAutoRedirect = false });
```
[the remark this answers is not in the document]
mjrousos: For configuring an HTTP client with cookies and redirect disabled? I don't know that's necessary; this is the only place we set up an HttpClient exactly like this.

src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/RemoteAppAuthenticationHttpHandler.Framework.cs
mjrousos: Combining them would work in most cases, but I worry about edge cases where some module after this one interacts with auth. I think it would be better to leave them separate but make sure all API key stuff happens in the module.

src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/RemoteAppAuthenticationOptions.Shared.cs
```diff
@@ -0,0 +1,69 @@
...
+
+    /// <summary>
+    /// Gets or sets the endpoint on the remote app that provides remote authentication
+    /// services. Requests to authenticate are sent to this endpoint.
+    /// </summary>
+    [Required]
+#endif
+    public string AuthenticationEndpointPath { get; set; } = "/systemweb-adapters/authenticate";
+
+#if NET6_0_OR_GREATER
```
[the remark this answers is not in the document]
mjrousos: Good idea. Fixed.

src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/RemoteAppAuthenticationResultFactory.cs
```diff
@@ -0,0 +1,59 @@
...
+        // If the remote authentication result hasn't yet been created, create it without a claims principal
+        if (ret is null)
+        {
+            ret = new RemoteAppAuthenticationResult(null, (int)response.StatusCode, response.RequestMessage);
+        }
+
+        // Copy expected response headers
+        foreach (var responseHeader in response.Headers)
+        {
+            if (!options.ResponseHeadersToForward.Any() || options.ResponseHeadersToForward.Contains(responseHeader.Key))
```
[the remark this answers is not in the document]
mjrousos: Sure. Fixed.

src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/RemoteAppAuthenticationService.cs
```diff
@@ -0,0 +1,120 @@
...
+/// Service for authenticating a user by making an HTTP request on their behalf to a remote app.
+/// </summary>
+internal partial class RemoteAppAuthenticationService : IRemoteAppAuthenticationService
+{
+    private bool _initialized;
+    private readonly HttpClient _client;
+    private readonly IAuthenticationResultFactory _resultFactory;
+    private readonly ILogger<RemoteAppAuthenticationService> _logger;
+    private readonly IOptionsMonitor<RemoteAppAuthenticationOptions> _optionsMonitor;
+    private RemoteAppAuthenticationOptions? _options;
```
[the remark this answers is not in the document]
mjrousos: Good point. I've made it an IOptionsSnapshot.

samples/RemoteAuth/Bearer/RemoteBearerCore/Program.cs
```diff
@@ -0,0 +1,31 @@
...
+
+var builder = WebApplication.CreateBuilder(args);
+builder.Services.AddReverseProxy().LoadFromConfig(builder.Configuration.GetSection("ReverseProxy"));
+
+// Add services to the container.
+
+builder.Services.AddControllers();
+
+builder.Services.AddSystemWebAdapters()
+    .AddRemoteAppAuthentication(true, options =>
```
davidfowl, line 11: What is the `true` parameter?

mjrousos: That's specifying whether remote app auth should be the default scheme or not. I updated the docs about that, but I'll add a comment here, as well.

src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/RemoteAppAuthenticationService.cs
```diff
@@ -0,0 +1,120 @@
...
+        using var response = await _client.SendAsync(authRequest, cancellationToken);
+        LogAuthenticationResponse(response.StatusCode);
+
+        return await _resultFactory.CreateRemoteAppAuthenticationResultAsync(response, _options);
+    }
+
+    // Add configured headers to the request, or all headers if none in particular are specified
+    private static void AddHeaders(IEnumerable<string> headersToForward, HttpRequest originalRequest, HttpRequestMessage authRequest)
+    {
+        // Add x-forwarded headers so that the authenticate API will
```
Tratcher, line 99: will what?

Tratcher (approved):

mjrousos:
> This is the main thing we need guidance for, we can't strand people in this hybrid state.

Agreed. Figuring out how to migrate off of forms auth is orthogonal to this PR, though, since this is just about bridging the gap so that the apps can share auth while endpoints migrate over gradually. Having a path forward for folks who used forms auth will be important, though, whether a user is taking advantage of these tools or just migrating all at once on their own.

twsouthwick (approved):

CZEMacLeod:
> > As far as how to migrate off of it, that's a good point. I think we'll probably want a follow-up feature that allows users to authenticate in the _ASP.NET Core_ app and forward the identity to the ASP.NET app (so, the reverse of what's happening here).
> 
> I don't know how many people will want that given they'll want to limit changes/investments in the old app. Have any customers we've interviewed expressed interest in that?
> 
> > I realize that doesn't make migrating off of forms auth any easier - that's a separate problem
> 
> This is the main thing we need guidance for, we can't strand people in this hybrid state.

I can state for a fact, that I would be interested in moving the auth to core, and forwarding to my asp.net app.
There are a number of improvements in core auth, and leaving it to last (e.g. move everything but auth to core, so that is all the asp.net app is doing then moving that) seems a bad way of progressing to me.

You already have to make changes to your old app (to handle session state etc.), so I don't see anything wrong with adding something else to provide auth sharing. I think there is already some stuff to handle this in `Microsoft.Owin.Security.Interop` with the `DataProtectorShim`, `AspNetTicketDataFormat`, and `ChunkingCookieManager` - although that probably only applies if you are using OWIN on both sides.

The only other thing I would say, is that being able to migrate my customized `Microsoft.AspNet.Identity.Core` / `Microsoft.AspNet.Identity.Owin` implementation without breaking my database format etc. would be a major help - but probably outside the scope of this project.

PR #41, opened 2022-05-24
Fix typo in property name

martincostello:
Change `HttpResponse` to `HttpRequest` for `UserLanguages`.

https://github.com/dotnet/systemweb-adapters/blob/8024bb69f1e10b9c8156b65f9f82cb1ed5fb3e0d/src/Microsoft.AspNetCore.SystemWebAdapters/HttpRequest.cs#L53

Also fixes two minor bits of casing in another document I spotted.


twsouthwick (approved):
LGTM - thanks for the review!

PR #44, opened 2022-05-24
Remove .NET Core 3.1 from support

### LINKED ISSUE #35: Consider dropping .NET Core 3.1 support
The adapter library currently supports targeting .NET Core 3.1. We should consider dropping support for .NET Core 3.1 since it will no longer be supported by the time we expect to ship a stable release.

twsouthwick:
Fixes #35


Tratcher (approved):
You can also remove it from https://github.com/dotnet/systemweb-adapters/blob/3244fac7fae0291bff97d2cc964abfee50fa71df/global.json#L9

PR #49, opened 2022-05-26
Call HttpResponse.FlushAsync instead of OutputStream.FlushAsync

twsouthwick:
A recent change changed things to call the flush on the stream rather than the request. This breaks the remote app session workflow as the output would not be flushed and so the core app would not be notified of the payload. Tests have been updated to verify this is called.

### DIFF
--- a/test/Microsoft.AspNetCore.SystemWebAdapters.SessionState.Framework.Tests/RemoteSession/GetWriteableSessionHandlerTests.cs
+++ b/test/Microsoft.AspNetCore.SystemWebAdapters.SessionState.Framework.Tests/RemoteSession/GetWriteableSessionHandlerTests.cs
@@ -88,6 +88,11 @@ public class GetWriteableSessionHandlerTests
         Assert.All(bytes.Skip(1), b => Assert.Equal((int)'\n', b));
 
         lockDisposable.Verify(d => d.Dispose(), Times.Once);
+
+        // Expect flush to be called each time a heartbeat is flushed. Since it will have already been called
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/RemoteSession/Framework/GetWriteableSessionHandler.cs
```diff
@@ -51,7 +51,9 @@ public async Task ProcessRequestAsync(HttpContextBase context, CancellationToken
 
         // Delimit the json body with a new line to mark the end of content
         context.Response.OutputStream.WriteByte(EndOfFrame);
-        await context.Response.OutputStream.FlushAsync(token);
+
+        // Ensure to call HttpResponse.FlushAsync to flush the request itself, and not context.Response.OutputStream.FlushAsync()
+        await context.Response.FlushAsync();
```
mjrousos, line 56: I think this was changed based on an analyzer run previously. Is there a diagnostic we need to suppress here or anything like that?

mjrousos (approved):

src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/RemoteSession/Framework/GetWriteableSessionHandler.cs
twsouthwick: I changed it figuring it was a similar API and thought it would be easier to test operating directly on stream. It was a last minute change that I hadn't run the actual E2E :( No analyzers for this that I'm aware of

PR #51, opened 2022-05-27
Break apart serialization of HttpSessionState properties and its keys

twsouthwick:
This breaks apart the two parts of serializing session state with different needs:

- HttpSessionState/ISessionState itself: This is probably not as interesting to implement for people as it is just some data to understand the session state. The abstraction ISessionSerializer is still used for this, but the key serialization has been removed. The new signature looks like this and is shared for .NET Core and .NET Framework:
  ```csharp
  public interface ISessionSerializer
  {
      Task<ISessionState?> DeserializeAsync(Stream stream, CancellationToken token);
  
      Task SerializeAsync(ISessionState state, Stream stream, CancellationToken token);
  }
  ```
- Keys and entries: This is the thing people are going to want to be able to customize. This change introduces an abstraction focused on this:
  ```csharp
  public interface ISessionKeySerializer
  {
      bool TrySerialize(string key, object value, out byte[] bytes);
  
      bool TryDeserialize(string key, byte[] bytes, out object? obj);
  }
  ```

With this change, someone who wants to really customize how session entries will be serialized can do whatever they want. They could chain together ISessionKeySerializer instances, or pick some strategy that makes sense for how they generate keys (i.e. dynamically on a pattern). Without this change, in order to do that, they would also have to implement the transformation for the whole HttpSessionState which is much more complicated.

Another nice thing about this structure, is that we can move majority of the session infrastructure into the core system adapters. An additional level would still be needed for System.Text.Json usage, but it could be a general package of all the JSON formatting people may want to opt into.

The final thing this change does is change the ISessionState a bit so that it can be available on .NET Framework as well without new dependencies. This is done by replacing IAsyncDisposable and ValueTask with IDisposable and Task. This makes implementation of some of the abstractions more straightforward and moves it to something that consumers can implement on .NET Standard.


src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/RemoteSession/RemoteAppSessionStateManager.cs
```diff
@@ -129,8 +125,13 @@ private async Task SetOrReleaseSessionData(ISessionState? state, CancellationTok
...
 
-            var bytes = _serializer.Serialize(state);
-            req.Content = new ByteArrayContent(bytes);
+            var ms = new MemoryStream();
+
+            await _serializer.SerializeAsync(state, ms, cancellationToken);
+
+            ms.Position = 0;
+
+            req.Content = new StreamContent(ms);
```
Tratcher, line 134: Perf: At some point we'll want to avoid creating this MemoryStream and buffering everything. Instead we'd create a custom HttpContent and serialized directly to the output stream.

src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/Serialization/BinarySessionSerializer.Shared.cs
```diff
@@ -0,0 +1,241 @@
...
+                writer.Write(key);
+            }
+        }
+
+        if (unknownKeys is not null && _options.ThrowOnUnknownSessionKey)
+        {
+            throw new UnknownSessionKeyException(unknownKeys);
+        }
+    }
+
```
Tratcher, line 111:
```suggestion
```

src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/Serialization/JsonSessionKeySerializer.Shared.cs
```diff
@@ -0,0 +1,64 @@
...
+        if (_options.KnownKeys.TryGetValue(key, out var type))
+        {
+            obj = JsonSerializer.Deserialize(bytes, type);
+            return true;
+        }
+
+        if (_options.ThrowOnUnknownSessionKey)
+        {
+            throw new UnknownSessionKeyException(key);
+        }
```
Tratcher, line 41: This option and exception seem out of place now that the method returns a bool. Methods using the Try pattern aren't supposed to throw. I'd expect the caller to check the return value and decide if it should throw or not.

Tratcher (approved):

src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/Serialization/JsonSessionKeySerializer.Shared.cs
twsouthwick: good point. I've moved that to the caller

src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/RemoteSession/RemoteAppSessionStateManager.cs
twsouthwick: Created an issue to track that

PR #58, opened 2022-06-01
Fix the session handler to catch operation canceled exceptions

mjrousos:
This fixes an issue in the GetWriteableSessionHandler that was causing shared session not to work properly. 

The (expected) OperationCanceledException when storing session state triggers the cancelation token that getting the session state is waiting on was causing ASP.NET's request processing to bail out and not successfully store updated session state.

This fix just catches the exception so that session updates are stored as expected on the ASP.NET side.

### DIFF
--- a/test/Microsoft.AspNetCore.SystemWebAdapters.SessionState.Framework.Tests/RemoteSession/GetWriteableSessionHandlerTests.cs
+++ b/test/Microsoft.AspNetCore.SystemWebAdapters.SessionState.Framework.Tests/RemoteSession/GetWriteableSessionHandlerTests.cs
@@ -76,7 +76,9 @@ public class GetWriteableSessionHandlerTests
         response.Verify(r => r.FlushAsync(), Times.Once);
 
         callback!();
-        await Assert.ThrowsAsync<TaskCanceledException>(() => task);
+
+        var exception = await Record.ExceptionAsync(() => task);
+        Assert.Null(exception);
 
         // Assert
         Assert.Equal(200, response.Object.StatusCode);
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.SystemWebAdapters.SessionState.Framework.Tests/RemoteSession/GetWriteableSessionHandlerTests.cs
```diff
@@ -133,7 +135,9 @@ public async Task DisconnectedRequest()
         response.Verify(r => r.FlushAsync(), Times.Once);
 
         cts.Cancel();
-        await Assert.ThrowsAsync<TaskCanceledException>(() => task);
+
+        var exception = await Record.ExceptionAsync(() => task);
```
twsouthwick, line 139: Were these throwing exceptions before?

mjrousos: They were (because of the cts.Cancel)

twsouthwick (approved):

test/Microsoft.AspNetCore.SystemWebAdapters.SessionState.Framework.Tests/RemoteSession/GetWriteableSessionHandlerTests.cs
mjrousos: Yep, looks like it. I see the exception if I turn on first chance exceptions in the debugger.

PR #68, opened 2022-06-09
Add HttpCookie.(Shareable|HasValues|Item)

### LINKED ISSUE #64: Add support for HttpCookie.Shareable
When constructing `HttpCookie`, the property `Shareable` is used to define whether something should be cached or not.  See https://referencesource.microsoft.com/#System.Web/HttpResponse.cs,ef49eacf1b7af1f6 for how it was used in ASP.NET.

twsouthwick:
Fixes #64


samples/MvcApp/MvcApp.csproj
```diff
@@ -283,7 +283,7 @@
           <AutoAssignPort>True</AutoAssignPort>
           <DevelopmentServerPort>50568</DevelopmentServerPort>
           <DevelopmentServerVPath>/</DevelopmentServerVPath>
-          <IISUrl>https://localhost:44339/</IISUrl>
+          <IISUrl>http://localhost:44339/</IISUrl>
```
Tratcher, line 286: ???

Tratcher (approved):

samples/MvcApp/MvcApp.csproj
twsouthwick: oops - didn't mean to commit that. That was part of verifying the behavior between ssl/non-ssl. I'll revert it

PR #76, opened 2022-06-24
Include a return URL in /Authenticate requests

### LINKED ISSUE #72: A 407 Proxy Authentication Required when using remote authentication
### Describe the bug
When using the remote authentication which was included in a recent update, after the authentication is complete, when the `/systemweb-adapters/authenticate` endpoint is loaded, an IIS HTTP 407.0 Proxy Authentication Required page is displayed.  After this error page is loaded, any attempt to go to any URL, whether it ends up loading the new ASP.NET Core application, or if it loads the old ASP.NET application, works without any further issues, with the user properly authenticated.  Also, the Logon User field on the error page does properly list the username of the person logged in.

This application is not using forms authentication nor bearer jwt.  It is instead making use of the `WsFederationAuthenticationModule` and `SessionAuthenticationModule` to allow federated login.  As I have it set up, the original ASP.NET application ran on localhost:44378, and the STS is set up to post the token to localhost:44378/, which would be processed by the `WsFederationAuthenticationModule`.  After the migration, the new ASP.NET Core application runs on localhost:44395.  When a challenge occurs, a redirect to the STS with the proper WS-Federation supplied information does occur, I am able to log in via the STS, and it posts the data to localhost:44378/, which properly processes the token, and then redirects to localhost:44378/systemweb-adapters/authenticate, at which point the Proxy Authentication Required page appears.  Attempting to reload localhost:44395 works with the logged in user information available.

I also attempted changing the STS to post the token information to localhost:44395/signin-wsfed (in order to make sure the request gets properly forwarded to the ASP.NET application rather than being handled by the default razor page in ASP.NET Core), the result is exactly the same, with the exception that the Proxy Authentication Required page is loaded from localhost:44395/systemweb-adapters/authenticate.  Also, just like before, reloading localhost:44395 after this results in proper authenticated access.

I do not have enough information to determine if this issue is caused by trying to use this with `WsFederationAuthenticationModule`, or if this is due to having something else not set up properly.

### To Reproduce
After performing a migration via the VS extension and then updating the SystemWebAdapters package to a locally built package containing the remote authentication implementation, I added code to enable the remote authentication.

ASP.NET Core application
```csharp
    .AddRemoteAppAuthentication(true, opts => {
        opts.RemoteServiceOptions.ApiKey = "Some secret key";
        opts.RemoteServiceOptions.RemoteAppUrl = 
            new(builder.Configuration["ReverseProxy:Clusters:fallbackCluster:Destinations:fallbackApp:Address"])
        ;
    })
```
ASP.NET Application
```csharp
    .AddProxySupport(opts => {
        opts.UseForwardedHeaders = true;
    })
    .AddRemoteAppAuthentication(opts => {
        opts.RemoteServiceOptions.ApiKey = "Some secret key";
    })
```

### Further technical details 

- Technologies and versions used (i.e. MVC/WebForms/etc): WebForms
- IIS Version: 10 Express
- Windows Version: 10

*ASP.NET Core Application*:
- Targeted .NET version: 6.0
- .NET SDK version: 6.0.300

mjrousos:
Some auth scenarios (WS-Fed, OIDC) were having issues with remote app auth because, as part of authentication, they redirected to an external auth provider to log the user in. When the auth provider redirected back to the app, it would re-direct back to the ASP.NET app's /systemweb-adapters/authenticate endpoint (which the ASP.NET Core app calls to in order to authenticate the user) instead of to the actual endpoint in the ASP.NET Core app that the user was originally trying to get to.

This PR fixes that by including the original request URL as a query string parameter when calling the authenticate endpoint (for example: https://localhost:4321/systemweb-adapters/authenticate?original-url=http://localhost:8765/SomeController/SomeAction (except that the original-url value is URL encoded)). Then, when the ASP.NET app handles request to /systemweb-adapters/authenticate, if no API key is provided (this is only present when called from the ASP.NET Core app), it will assume that the request may be a redirect back and if an original-url query string is present (the auth flows I tested all preserved the query string), the ASP.NET app will re-direct the request to that URL instead.

This PR also adds an OIDC sample to demonstrate this scenario working.

Fixes #72


src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/RemoteAppAuthenticationModule.Framework.cs
```diff
@@ -28,10 +28,36 @@ void MapRemoteAuthenticationHandler(object sender, EventArgs e)
             if (string.Equals(context.Request.Path, _options.AuthenticationEndpointPath, StringComparison.OrdinalIgnoreCase)
                 && context.Request.HttpMethod.Equals("GET", StringComparison.OrdinalIgnoreCase))
             {
-                if (!string.Equals(_options.RemoteServiceOptions.ApiKey, context.Request.Headers.Get(_options.RemoteServiceOptions.ApiKeyHeader), StringComparison.Ordinal))
+                var apiKey = context.Request.Headers.Get(_options.RemoteServiceOptions.ApiKeyHeader);
+                if (apiKey is null)
```
twsouthwick, line 32: Can you add some tests around this? This logic should be safeguarded behind unit tests so we make sure not to regress

mjrousos: Yep, I agree we should have some tests. I'm out next week, but will get this updated with unit tests when I'm back after the 4th.

mjrousos: Done

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/ProxyHeaderModule.Framework.cs
```diff
@@ -34,12 +34,53 @@ public void Init(HttpApplication context)
 
             context.BeginRequest += (s, e) => UseOptions(((HttpApplication)s).Context.Request);
         }
+
+        context.PreSendRequestHeaders += (s, e) => FixupLocationHeader(((HttpApplication)s).Context);
```
twsouthwick, line 38: I wasn't familiar with this event, but the docs have this dire warning that I think we would be included in: https://docs.microsoft.com/en-us/dotnet/api/system.web.httpapplication.presendrequestheaders?view=netframework-4.8#remarks

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/ProxyHeaderModule.Framework.cs
```diff
@@ -59,11 +102,11 @@ private static void UseHeaders(HttpRequest request)
             {
                 request.ServerVariables.Set("SERVER_PORT", value.Port);
             }
-        }
 
-        if (request.Headers["x-forwarded-proto"] is { } proto)
-        {
-            request.ServerVariables.Set("SERVER_PROTOCOL", proto);
+            if (request.Headers["x-forwarded-proto"] is { } proto)
```
twsouthwick, line 106: do we not want to set the proto if the host is not available?

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/ProxyHeaderModule.Framework.cs
```diff
@@ -75,4 +118,26 @@ private static void UseForwardedFor(HttpRequest request)
...
+    }
+
+    private static int? GetPort(string? portString, string protocol)
+    {
+        if (int.TryParse(portString, out var port))
+        {
+            return port;
+        }
+
+        return protocol.ToUpperInvariant() switch
```
twsouthwick, line 136: nit: this would do a string allocation every request

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/ProxyHeaderModule.Framework.cs
```diff
@@ -34,12 +34,53 @@ public void Init(HttpApplication context)
 
             context.BeginRequest += (s, e) => UseOptions(((HttpApplication)s).Context.Request);
         }
+
+        context.PreSendRequestHeaders += (s, e) => FixupLocationHeader(((HttpApplication)s).Context);
+    }
+
+    private static void FixupLocationHeader(HttpContext context)
+    {
+        // Some scenarios (such as OIDC auth) will store a URL (including the local host/port rather than the forwarded one)
```
twsouthwick, line 43: where is the OIDC getting the information about host/port? AFAIK part of setting the proxy stuff is so that things were automatically accounted for

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/ProxyHeaderModule.Framework.cs
```diff
@@ -34,12 +34,53 @@ public void Init(HttpApplication context)
...
+        // and later redirect users to that URL. If a response contains a Location header, ensure that the correct scheme,
+        // host, and port are used.
+        if (context.Response.Headers["Location"] is { } location)
```
twsouthwick, line 46: why include this fix up here rather than in the remote app module?

src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/RemoteAppAuthenticationModule.Framework.cs
```diff
@@ -2,44 +2,96 @@
...
+        {
+            // A missing API key may indicate that instead of being a request from the ASP.NET Core app,
+            // this request represents a redirect back from an external authentication provider (for example,
+            // in OIDC or WS-Fed scenarios). In those scenarios, the external authentication provider may attempt
+            // to redirect back to the URL that initiated the login request.
+            //
+            // In those scenarios, if there is an original-url query parameter present in the request's URL, we
+            // can redirect back to that as the *actual* original URL that the authentication process needs to
+            // redirect back to.
+            var originalUrl = context.Request.QueryString[AuthenticationConstants.OriginalUrlQueryParamName];
```
twsouthwick, line 72: do we need to do any sort of data protection on this? i.e. could this be used in a nefarious way?

twsouthwick (changes requested):

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/ProxyHeaderModule.Framework.cs
mjrousos: It seemed like a weird scenario to set a different proto without changing host or port. Is there any scenario where we would forward from http to https or vice-versa with the same host and port?

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/ProxyHeaderModule.Framework.cs
mjrousos: Good point. I'll just do a couple invariant comparisons instead.

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/ProxyHeaderModule.Framework.cs
mjrousos: I haven't dug in to see exactly where it's getting them from, but I found that when a challenge results in redirecting to OIDC login, the URL that the user was trying to navigate to gets stored in a serialized state blob that travels along for the OIDC journey and, at the end, gets spit back out in a Location header to redirect the user to that URL. Somehow the OIDC code that handles challenges was picking up the ASP.NET host/port instead of the ASP.NET Core one even with SERVER_NAME and SERVER_PORT set.

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/ProxyHeaderModule.Framework.cs
mjrousos: I could see it living either place. If OIDC code worked the way I wanted it to, this would have been handled by `UseHeaders` in this module (by setting SERVER_NAME, etc.). Because that doesn't work, we have the 'backup plan' of checking Location headers here.

We could put this in the remote app authentication module since auth scenarios are the ones we know of now that lead to incorrect Location response headers, but if there were ever any other buggy modules/handlers/etc. that resulted in Location headers pointing at the old app we'd want to fix those up, too.

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/ProxyHeaderModule.Framework.cs
mjrousos: Ok, I figured out why this is necessary.

The problematic code is in the OWIN OpenIdConnectAuthenticationHandler: https://github.com/aspnet/AspNetKatana/blob/main/src/Microsoft.Owin.Security.OpenIdConnect/OpenidConnectAuthenticationHandler.cs#L137

When auth is challenged the OWIN OIDC handler creates a redirect URI based on its Request property (of type Microsoft.Owin.OwinRequest). That property has a "Host" member which is not affected by the ServerVariables set on the HttpRequestBase earlier (even though I can see the x-forward headers in the OwinRequest!).

@Tratcher - any ideas how we can fix this on the OWIN side? The summary is that OpenIdConnectAuthenticationHandler is looking at the current URI to figure out where to redirect users back to after auth in challenge scenarios and is coming up with a URI that contains the ASP.NET host/port instead of the ASP.NET Core one even though x-forward headers are set and request ServerVariables like SERVER_NAME and SERVER_PORT have been set. This PR works around the issue by adding some code that watches response headers coming from the ASP.NET app and fixes up any Location headers trying to redirect the user to the old app instead of the new one. Still, if OWIN used the new redirect URI here that workaround wouldn't be necessary.

image

twsouthwick: I think I found a fix. I was fixing a different bug with the proxy stuff, and we aren't changing the Host header. Owin is using that to grab this info. I've update the module in #92 to handle that (it adds a bunch of tests around that as well, which will slightly conflict with the tests you've added here).

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/ProxyHeaderModule.Framework.cs
twsouthwick: See above. I believe this is fixed in #92

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/ProxyHeaderModule.Framework.cs
mjrousos: Awesome! I'll test that out and if that addresses the issue with OIDC scenarios, then I can revert that part of this PR.

twsouthwick: It's merged into main now, so hopefully that fixes this issue

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/ProxyHeaderModule.Framework.cs
mjrousos: Awesome. I'll see if I can revert this part of the PR then.

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/ProxyHeaderModule.Framework.cs
mjrousos: Removed since this is no loner necessary.

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/ProxyHeaderModule.Framework.cs
mjrousos: I've confirmed that #92 fixes the issue with OIDC not respecting forwarded host/port, so I've reverted this part of the PR.

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/ProxyHeaderModule.Framework.cs
mjrousos: Reverted since I'm no longer touching this class for any other reason and it doesn't seem important enough to change on its own.

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/ProxyHeaderModule.Framework.cs
mjrousos: I've confirmed that https://github.com/dotnet/systemweb-adapters/pull/92 fixes the issue with OIDC not respecting forwarded host/port, so I've reverted this part of the PR.

src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/RemoteAppAuthenticationModule.Framework.cs
twsouthwick: Hmm that makes sense. It wouldn't end up bypassing anything or disclosing any information, right? It's probably fine then

src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/ResultProcessors/RedirectUrlProcessor.cs
```diff
@@ -7,41 +7,38 @@
...
     /// </summary>
     /// <param name="result">The remote app authentication result to be updated so that redirects point to the right locations.</param>
     /// <param name="context">The HTTP context containing the original request that prompted authentication.</param>
     public Task ProcessAsync(RemoteAppAuthenticationResult result, HttpContext context)
     {
         if (result.ResponseHeaders.TryGetValue(LocationHeaderName, out var locationHeaders))
         {
-            // Look for any Location headers with URLs including a ReturnUrl query string as their value
+            // Look for any Location headers
             var processedRedirectLocations = new List<string>();
```
twsouthwick, line 33: should we defer creating the list until we know it's needed? How often will there be updated locations, but no values?

test/Microsoft.AspNetCore.SystemWebAdapters.Framework.Tests/Microsoft.AspNetCore.SystemWebAdapters.Framework.Tests.csproj
```diff
@@ -1,7 +1,8 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net48</TargetFrameworks>
+    <TargetFramework>net472</TargetFramework>
```
twsouthwick, line 4: why this change?

test/Microsoft.AspNetCore.SystemWebAdapters.Framework.Tests/ProxyHeaderModuleTests.cs
```diff
@@ -4,7 +4,7 @@
 using System.Collections.Specialized;
 using Xunit;
 
-namespace Microsoft.AspNetCore.SystemWebAdapters.FrameworkTests;
+namespace Microsoft.AspNetCore.SystemWebAdapters.Tests;
```
twsouthwick, line 7: :)

twsouthwick (approved):
LGTM. I think we'll want @tratcher to review as well.

test/Microsoft.AspNetCore.SystemWebAdapters.Framework.Tests/ProxyHeaderModuleTests.cs
mjrousos: Made it match some of the other framework tests

test/Microsoft.AspNetCore.SystemWebAdapters.Framework.Tests/Microsoft.AspNetCore.SystemWebAdapters.Framework.Tests.csproj
mjrousos: This came in when I merged my Microsoft.AspNetCore.SystemWebAdapters.Framework.Tests.csproj with what you added. I had chose 472 just because that's what test\Microsoft.AspNetCore.SystemWebAdapters.SessionState.Framework.Tests\Microsoft.AspNetCore.SystemWebAdapters.SessionState.Framework.Tests.csproj was using and I figured they may as well match. Either version is fine.

src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/RemoteAppAuthenticationModule.Framework.cs
mjrousos: Right

test/Microsoft.AspNetCore.SystemWebAdapters.Framework.Tests/Microsoft.AspNetCore.SystemWebAdapters.Framework.Tests.csproj
mjrousos: Yea, that's what I gathered (though I think it's arbitrary).

src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/ResultProcessors/RedirectUrlProcessor.cs
mjrousos: It should almost always be a single string. I'm looping through all values here since headers, generally, can have multiple values but I can't think of any case where it would make sense to have more than one Location header.

samples/RemoteAuth/OIDC/OIDCAuth/Controllers/AccountController.cs
```diff
@@ -0,0 +1,53 @@
...
+
+            return View();
+        }
+
+        private static string GetHost(HttpRequestBase request) =>
+            request.Headers["x-forwarded-host"]
+                ?? (request.Url.IsDefaultPort ? request.Url.Host : $"{request.Url.Host}:{request.Url.Port}");
+
+        private static string GetScheme(HttpRequestBase request) =>
+            request.Headers["x-forwarded-proto"] ?? request.Url.Scheme;
```
Tratcher, line 51: Did we decide if it's possible for our module to automatically apply these headers to the request so you don't have to customize application code like this?

Tratcher (changes requested):

samples/RemoteAuth/OIDC/OIDCAuth/Controllers/AccountController.cs
mjrousos: Good catch. I think we have modules to do this now; I'll update this.

test/Microsoft.AspNetCore.SystemWebAdapters.Framework.Tests/Microsoft.AspNetCore.SystemWebAdapters.Framework.Tests.csproj
twsouthwick: I believe 4.7.2 is the recommended target because that's the version that async context/locals (and maybe some other things more modern code relies on) works more aligned with core. Not sure if that affects what we're doing though

src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/RemoteAppAuthenticationModule.Framework.cs
```diff
@@ -2,44 +2,96 @@
...
-                }
             }
+        };
+    }
+
+    public void MapRemoteAuthenticationHandler(HttpContextBase context)
+    {
+        var apiKey = context.Request.Headers.Get(_options.RemoteServiceOptions.ApiKeyHeader);
+        if (apiKey is null)
+        {
```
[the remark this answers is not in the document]
mjrousos: This should be better now

twsouthwick (approved):
Ooh this seems much more straightforward. LGTM!

samples/RemoteAuth/OIDC/OIDCAuth/Controllers/AccountController.cs
twsouthwick: Yeah, I think this is handled in the proxy module with a recent change I did.

mjrousos: Yep, I've confirmed this isn't needed anymore, so I've removed `GetHost` and `GetScheme` from here.

Tratcher (approved):

mjrousos:
While testing this, I found that the OWIN property approach worked well for OIDC but broke Identity/cookie auth scenarios. In OIDC scenarios, the .redirect property was used to know which URL to re-direct the user back to after authentication completes. However, in cookie scenarios, the .redirect property is being used to determine which URL to re-direct the user to in order to authenticate!

The end-result is that the OWIN property-based change that fixed the OIDC scenario broke cookie scenarios because the URL the user was ultimately trying to get to was set as .redirect and when the OWIN cookie auth handler tried to authenticate the user, it would send them to _that_ URL to sign in causing an infinite loop.

I spent a while trying to figure out if there was a clean way of determining in our HTTP handler which OWIN auth handler was being used (OIDC, cookie, or something else) so that we could change what we set .redirect to accordingly but I couldn't find a good way to do that.

Consequently, I've reverted back to an early approach where the relative path that the user is trying to reach is included as a query string in the ASP.NET Core's call to /authenticate to determine user identity. The ASP.NET Core app also includes a special header indicating that the request is from the ASP.NET Core app for purposes of authentication. All of the OWIN auth flows I've tested will eventually redirect back to the URL where authentication began (/authenticate in this case) and will preserve the query strings but not the headers. This allows the /authenticate HTTP handler to recognize when it's being called as part of a redirect in the auth process (the special "I'm authenticating" header will be missing). In those situations, it redirects back to the path in the query string (relative to the ASP.NET Core app's host) and the user is returned to the path they expected to be at after authenticating.

I've already chatted offline with @Tratcher about this approach, but given that the code is different from how it was approved previously, I'd appreciate another round of reviews, @Tratcher and @twsouthwick. Thanks!

twsouthwick (approved):
This seems fine. I'm glad it's now a relative path - that seems a lot better.

src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/RemoteAppAuthenticationModule.Framework.cs
```diff
@@ -2,47 +2,112 @@
...
+            }
+            else
+            {
+                // A request without a migration authentication header and without an original URL to redirect
+                // back to is invalid. Return 400 to indicate that the request was malformed.
+                context.Response.StatusCode = 400;
+            }
+
+            // Clear any existing handler as this request is now completely handled
+            context.Handler = null;
```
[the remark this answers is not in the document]
mjrousos: It's not necessarily null. This is run post mapping request handlers and I noticed while testing the scenario that this was typically set to an MVC handler (so if this module doesn't change anything, the MVC handler will attempt to handle the request).

There isn't a controller for /systemweb-adapters/authenticate, of course, so the MVC handler was overwriting the 302 or 400 or whatever that this module set with a 401.

I think this may not have been necessary previously when the module always set the handler for requests to the authenticate path. But now that the module is doing some of the checks for whether to handle the request or not, we need to clear the handler in cases that we aren't explicitly setting it (and don't want to fall back to whatever the default would be like MVC).

src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/RemoteAppAuthenticationModule.Framework.cs
```diff
@@ -2,47 +2,112 @@
...
+            //
+            // This most likely indicates that an identity provider is redirecting back to the application after
+            // authenticating the user. In that case, the original-url query string will indicate the path
+            // that the user should be redirected back to.
+            var originalUrlPath = context.Request.QueryString[AuthenticationConstants.OriginalUrlQueryParamName];
+            if (originalUrlPath is not null)
+            {
+                context.Response.StatusCode = 302;
+
+                // Redirect back to the provider path relative to the current host
```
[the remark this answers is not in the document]
mjrousos: I've updated the comment to try and make it clearer.

src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/RemoteAppAuthenticationModule.Framework.cs
```diff
@@ -2,47 +2,118 @@
...
+
+                // Redirect back to the provided path relative to the host used to
+                // access this endpoint (the host from the current request's URL).
+                //
+                // Because external identity providers will call back to this endpoint
+                // via the ASP.NET Core app (which then proxies to here), this will
+                // create a redirect that uses the original request's path with the
+                // ASP.NET Core app's host/port.
+                var redirect = new Uri(context.Request.Url, originalUrlPath);
+                context.Response.Headers["Location"] = redirect.ToString();
```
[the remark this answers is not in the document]
mjrousos: Updated

PR #78, opened 2022-06-28
Set current principal if available

twsouthwick:


twsouthwick:
@davidfowl  @Tratcher Is there a reason we shouldn't do this? The backing of the `Thread.CurrentPrincipal` is an async local so it should be handled fine and allow code written against that to *just work*. It also may be worth considering just adding it into ASP.NET Core itself.

Tratcher:
@blowdart

davidfowl:
I think this is fine from a compat perspective but not from a "do by default in ASP.NET Core" perspective:
- We've moved away from encouraging people using `Thread.CurrentPrincipal`. If we were going to do something like this, it would model IHttpContextAccessor.
- I've recently come to realize that having anything tied to the HttpContext on an async local is badness. There's no way to control the lifetime and how this thing flows. That can lead to concurrent access when it isn't supported.
- Performance, the more async locals the worse the performance. We try to avoid having more by default (we already have 2 by default).

blowdart:
Yea, thread.currentprincipal is to be avoided as much as possible, it messes up badly with async as David says, it never worked well with impersonation (not that we're encouraging impersonation either), and 1st and 3rd party libraries made so many bad assumptions we pushed the AAD folks away from it. 

I'm not sure we would even want this from a compat POV given its tendency to go wrong.

twsouthwick:
@blowdart @davidfowl Thanks for the comments

I've updated this PR to do the following:

- Add metadata to be used to enable per-endpoint setting the Thread.CurrentPrincipal (which will then surface through ClaimsPrincipal.Current as well)
- A warning will be logged if this is set - it is currently a Warning, but that may end up being too noisy. Not sure if there's a better way, though
- Check if the logical single thread mode is enabled for the request. If not, it will log a warning that that should be enabled
- Add metadata to be used to enable per-endpoint logical single-threaded behavior using ConcurrentExclusiveSchedulerPair

I did also try overwriting the ClaimsPrincipal.Selector function to log warnings if it is accessed and to use HttpContextAccessor directly, but ended up removing it as the other warnings will probably be sufficient.

This is all opt-in and hopefully will help users understand it's not best practice but enable them to more easily migrate old code.

davidfowl:
We shouldn’t have a wanting that logs per request, let’s get rid of it. Put something in the doc comments.

docs/usage_guidance.md
```diff
@@ -35,12 +29,26 @@ Simplest way to enable this with similar behavior as ASP.NET Framework would be
 app.UseRequestLocalization();
 ```
 
+## `System.Threading.Thread.CurrentPrincipal`
+
+In ASP.NET Framework, `System.Threading.Thread.CurrentPrincipal` and `System.Security.Claims.ClaimsPrincipal.Current` would be set to the current user. This is not available on ASP.NET Core out of the box. Support for this is available with these adapters by adding the `ICurrentPrincipalMetadata` to the endpoint. However, it should only be used if the code cannot be refactored to remove usage.
```
Tratcher, line 34: Mention the attribute too.

Tratcher (approved):

PR #81, opened 2022-06-29
Add HttpRuntime.AppDomainAppVirtualPath

twsouthwick:


CZEMacLeod:
Does this lead to being able to create a version of `System.Web.VirtualPathUtility` that works under ASP.Net Core?
I think I only really need 
```cs
public interface IVirtualPathUtility
{
	string ToAbsolute(string virtualPath);
	string ToAppRelative(string virtualPath);
}
public class VirtualPathUtility : IVirtualPathUtility
{
	public string ToAbsolute(string virtualPath) => System.Web.VirtualPathUtility.ToAbsolute(virtualPath);
	public string ToAppRelative(string virtualPath) => System.Web.VirtualPathUtility.ToAppRelative(virtualPath);
}
```

twsouthwick:
@CZEMacLeod this is probably part of it. Can you open an issue with the APIs you would need? The whole VirtualPath infrastructure ends up pulling in a bit of stuff, so we're initially just going after APIs that are used in the wild.

Tratcher (approved):

PR #82, opened 2022-06-29
Add middleware to restrict concurrency

twsouthwick:
This adds middleware that will turn a request into (logically) a single-threaded request. This helps mimic some of the behavior that was seen on ASP.NET Framework where there was a concept of a request thread. Since ASP.NET Core is not tied to a specific thread, that model does not translate directly. However, by enabling this, it will limit concurrent access to items within the flow of a request.


twsouthwick:
Closing as #78 is incorporating this

PR #84, opened 2022-06-30
Separate out JSON session from others so it can be overridden

twsouthwick:


src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/RemoteSession/Framework/RemoteAppSessionStateExtensions.cs
```diff
@@ -9,7 +9,7 @@ namespace Microsoft.AspNetCore.SystemWebAdapters;
 
 public static class RemoteAppSessionStateExtensions
 {
-    public static ISystemWebAdapterBuilder AddRemoteAppSession(this ISystemWebAdapterBuilder builder, Action<RemoteAppSessionStateOptions> configureRemote, Action<JsonSessionSerializerOptions> configureSerializer)
+    public static ISystemWebAdapterBuilder AddRemoteAppSession(this ISystemWebAdapterBuilder builder, Action<RemoteAppSessionStateOptions> configureRemote, ISessionKeySerializer keySerializer)
```
twsouthwick, line 12: This API is new and allows a user to customize the session key serialization. This was already able to be customized on the core side because of DI, but on the framework side, we aren't using that.

Issue #91, opened 2022-07-07
Add some missing APIs for HttpContext, HttpRequest and HttpResponse


Maples7:
## Summary

Most primary APIs of HttpContext, HttpRequest and HttpResponse are provided, which is fantastic! And some missing APIs could be considered to be added to make this adapter more universal.

## Motivation and goals

Some missing APIs are frequently used for .NET Framework and adding them to the adapter will make the migration smoother and more doable for large projects.

## In scope

For HttpContext:
- [x] DisposeOnPipeline https://github.com/dotnet/systemweb-adapters/pull/88
- [ ] CurrentHandler https://github.com/dotnet/systemweb-adapters/issues/146
- [ ] CurrentNotification https://github.com/dotnet/systemweb-adapters/issues/146
- [ ] IsCustomErrorEnabled https://github.com/dotnet/systemweb-adapters/issues/152
- [x] IsDebuggingEnabled https://github.com/dotnet/systemweb-adapters/pull/106
- [ ] PreviousHandler https://github.com/dotnet/systemweb-adapters/issues/146
- [x] Timestamp https://github.com/dotnet/systemweb-adapters/pull/104
- [ ] GetGlobalResourceObject https://github.com/dotnet/systemweb-adapters/issues/23
- [ ] GetLocalResourceObject https://github.com/dotnet/systemweb-adapters/issues/23
- [ ] GetSection https://github.com/dotnet/systemweb-adapters/issues/23
- [ ] RewritePath https://github.com/dotnet/systemweb-adapters/pull/145
- [ ] SkipAuthorization https://github.com/dotnet/systemweb-adapters/issues/147
- [ ] SetSessionStateBehavior https://github.com/dotnet/systemweb-adapters/issues/147

For HttpRequest:
- [x] AcceptTypes https://github.com/dotnet/systemweb-adapters/pull/101
- [ ] ClientCertificate https://github.com/dotnet/systemweb-adapters/issues/148
- [ ] FilePath https://github.com/dotnet/systemweb-adapters/issues/149
- [x] Files https://github.com/dotnet/systemweb-adapters/pull/103
- [x] Item[String] https://github.com/dotnet/systemweb-adapters/pull/105
- [x] Params https://github.com/dotnet/systemweb-adapters/pull/105
- [ ] PathInfo https://github.com/dotnet/systemweb-adapters/issues/149
- [ ] PhysicalPath https://github.com/dotnet/systemweb-adapters/issues/149
[...]

PR #102, opened 2022-07-11
Add HttpResponse APIs (ClearHeaders/WriteFile/TransmitFile/etc)

twsouthwick:
This change adds a few APIs that were missing from HttpResponse.

SubStatusCode is a placeholder for the value, but does not actually publish it anywhere.

TransmitFile/WriteFile: These APIs are essentially the same, except that they had different memory characteristics. TransmitFile did not buffer, while WriteFile did. On ASP.NET Core, these will just delegate to IHttpResponseBodyFeature.SendFileAsync.

NOTE: This implementation will call Task.GetAwaiter() which should be avoided. We could potentially expose an async API for this on the ASP.NET Core side as a stepping stone if we want.

Part of #91


src/Microsoft.AspNetCore.SystemWebAdapters/HttpResponse.cs
```diff
@@ -45,14 +45,29 @@ public int StatusCode
             set => _response.StatusCode = value;
         }
 
-        public string StatusDescription
+        public int SubStatusCode { get; set; } = -1;
```
mjrousos, line 48: Why -1? It looks like this defaulted to 0 in ASP.NET

mjrousos (changes requested):
Looks good other than a question about SubStatusCode's initial value

src/Microsoft.AspNetCore.SystemWebAdapters/HttpResponse.cs
twsouthwick: good catch. there was a different value that had a default as -1 and I think I got that mixed up

mjrousos (approved):
LGTM

src/Microsoft.AspNetCore.SystemWebAdapters/HttpResponse.cs
```diff
@@ -167,6 +182,26 @@ public void AppendHeader(string name, string value)
             }
         }
 
+        public bool IsRequestBeingRedirected { get; private set; }
```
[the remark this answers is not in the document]
twsouthwick: done

PR #103, opened 2022-07-11
Add HttpRequest.Files and supporting types

twsouthwick:
Part of #91


src/Microsoft.AspNetCore.SystemWebAdapters/HttpFileCollection.cs
```diff
@@ -0,0 +1,147 @@
...
+
+    [SuppressMessage("Performance", "CA1819:Properties should not return arrays", Justification = Constants.ApiFromAspNet)]
+    public string[] AllKeys => _keys ??= GetKeys();
+
+    public override int Count => FormFiles.Count;
+
+    [Obsolete("Retrieving Keys is not supported on .NET 6+. Please use the enumerator instead.")]
+#pragma warning disable CS0809 // Obsolete member overrides non-obsolete member
+    public override KeysCollection Keys => throw new PlatformNotSupportedException();
+#pragma warning restore CS0809 // Obsolete member overrides non-obsolete member
```
Tratcher, line 34: How is this different from AllKeys?

src/Microsoft.AspNetCore.SystemWebAdapters/HttpFileCollection.cs
```diff
@@ -0,0 +1,147 @@
...
+
+    public override IEnumerator GetEnumerator() => AllKeys.GetEnumerator();
```
Tratcher, line 36: Should this really be returning keys rather than files?

src/Microsoft.AspNetCore.SystemWebAdapters/HttpPostedFile.cs
```diff
@@ -0,0 +1,22 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.IO;
+using Microsoft.AspNetCore.Http;
+
+namespace System.Web;
+
+public sealed class HttpPostedFile
+{
+    public HttpPostedFile(IFormFile file) => File = file;
```
Tratcher, line 11: Should this constructor be internal?

Tratcher (approved):

src/Microsoft.AspNetCore.SystemWebAdapters/HttpFileCollection.cs
twsouthwick: KeysCollection ends up using the int indexer and there's no way to override that behavior.

src/Microsoft.AspNetCore.SystemWebAdapters/HttpFileCollection.cs
twsouthwick: Weirdly, that's what the NameObjectCollectionBase does. You iterate through to get the keys, and then you end up having to index into it by string

src/Microsoft.AspNetCore.SystemWebAdapters/HttpPostedFile.cs
twsouthwick: it's only available on .NET 6 and I did that so people could test things if they wanted. I can make it internal if you have a preference (we test it, but we have internals visible to)

PR #106, opened 2022-07-12
Add HttpContext.IsDebuggingEnabled

twsouthwick:
On ASP.NET, this was tied to the compilation model used for controls. Since this doesn't exist on framework, it makes most sense to expose whether the current environment is development.

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters/HttpContext.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters/HttpContext.cs
@@ -7,9 +7,11 @@ using System.Security.Claims;
 using System.Security.Principal;
 using System.Web.Caching;
 using System.Web.SessionState;
+using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.SystemWebAdapters;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
 
 namespace System.Web;
 
@@ -40,6 +42,11 @@ public class HttpContext : IServiceProvider
     public HttpServerUtility Server => _server ??= new(_context);
 
     public Cache Cache => _context.RequestServices.GetRequiredService<Cache>();
+
+    /// <summary>
+    /// Gets whether the current request is running in the development environment.
+    /// </summary>
+    public bool IsDebuggingEnabled => _context.RequestServices.GetRequiredService<IWebHostEnvironment>().IsDevelopment();
 
     public IPrincipal User
     {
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.SystemWebAdapters/HttpContext.cs
```diff
@@ -41,6 +43,11 @@ public HttpContext(HttpContextCore context)
 
     public Cache Cache => _context.RequestServices.GetRequiredService<Cache>();
 
+    /// <summary>
+    /// Gets whether the current request is running in the development environment.
+    /// </summary>
+    public bool IsDebuggingEnabled => _context.RequestServices.GetRequiredService<IWebHostEnvironment>().IsDevelopment();
```
davidfowl, line 49: This makes me uncomfortable but I see why it's written this way.

twsouthwick: Which part makes you uncomfortable?

mjrousos (approved):

src/Microsoft.AspNetCore.SystemWebAdapters/HttpContext.cs
twsouthwick: The way that this value was set in ASP.NET Framework is in the web.config:

```xml
  <system.web>
    <compilation debug="true" targetFramework="4.7.2">
      <assemblies>
        <add assembly="netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51" />
      </assemblies>
    </compilation>
    <httpRuntime targetFramework="4.7.2" />
  </system.web>
```

there's a web.config transform that occurs on publish that removes the debug (see `Web.Release.config`):

```xml
 <system.web>
    <compilation xdt:Transform="RemoveAttributes(debug)" />
  </system.web>
```

Given that it was a static value that was mostly on dev time, having it based on the WebHostEnvironment made the most sense.

We could also consider abstracting it in some way so someone could customize it, but at this point that seems like too much effort for payoff.

mjrousos: `IWebHostEnvironment.IsDevelopment()` feels closer to me because `IsDebuggingEnabled` should return true if the request is being served in debug mode, not just if there's a debugger attached to the process.

davidfowl: Right, I looked and I agree. I was also going to suggest `IsDebuggingEnabled` but realized we were looking at the compilation section on .NET Framework.

PR #110, opened 2022-07-14
Enable configuring remote app options directly (separate from auth or session)

### LINKED ISSUE #50: Simplify registration of remote app urls
For things like remote app session, we require setting up the remote app. This is currently being done by digging into the same config value that YARP is using:

```csharp
    .AddRemoteAppSession(options =>
    {
        // Provide the URL for the remote app that has enabled session querying
        options.RemoteApp = new(builder.Configuration["ReverseProxy:Clusters:fallbackCluster:Destinations:fallbackApp:Address"]);

        // Provide a strong API key that will be used to authenticate the request on the remote app for querying the session
        options.ApiKey = "strong-api-key";
    });
```

This looks a little ugly - can we refine this at all?

### LINKED ISSUE #56: Consider adding MS.Extensions.DependencyInjection and MS.Extensions.Options for framework adapters
## Summary

We should add the Microsoft.Extensions.DependencyInjection and Microsoft.Extensions.Options packages to the framework app so that configuration between the two apps can be similar.

## Motivation and goals

There's currently two separate mechanisms for configuring the core app and the framework app. The core app uses the standard dependency injection patterns as the rest of ASP.NET Core. The framework app uses a similar fluent syntax, but doesn't have the full dependency injection available to it. This will make it difficult for users to fully customize things, such as custom session serializers.

As customers move to ASP.NET Core, they will start taking advantage of using the new DI and options frameworks, and this can be a way to simplify things in their ASP.NET Framework code base. This will also allow the adapters to be more flexible and have a consistent usage across app stacks.

## In scope

- Main use will be to update `ISystemWebBuilder` to expose `IServiceCollection` rather than `ICollection<IHttpModule>` on framework. No changes will be seen to the core adapters.

## Out of scope

- If we want to expose `ISystemWebBuilder` on .NET Standard, we could just add the abstractions packages. However, I'm leaning towards not exposing it to reduce that dependency if someone is targeting .NET Standard
 
## Risks / unknowns

- This may bring in new dependencies to the users framework app.

mjrousos:
This allows users to configure remote app endpoint and API key information once and use those options for both auth and session (and any future features that require that information).

Also, adds DI to make dependency management easier in System.Web modules.

Fixes #50 
Fixes #56


src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/RemoteAppExtensions.Shared.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            throw new ArgumentNullException(nameof(configure));
+        }
+
+        var options = builder.Services.AddOptions<RemoteAppOptions>()
+            .Configure(configure);
+
+#if NET6_0_OR_GREATER
+        options.ValidateDataAnnotations();
+#endif
```
twsouthwick, line 31:
```suggestion
        builder.Services.AddOptions<RemoteAppOptions>()
#if NET6_0_OR_GREATER
            .ValidateDataAnnotations();
#endif
            .Configure(configure);

```

twsouthwick (approved):
Thanks @mjrousos. This will make a number of the patterns simpler for both apps. Just a couple of comments that I don't have very strong opinions about.

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/RemoteAppExtensions.Shared.cs
mjrousos: Updated (though I didn't take the suggestion directly since there's an extra ; in it).

src/Microsoft.AspNetCore.SystemWebAdapters/Microsoft.AspNetCore.SystemWebAdapters.csproj
```diff
@@ -32,6 +32,9 @@
     <Compile Remove="**/*" />
     <Compile Include="@(StandardFiles)" />
     <Compile Include="@(SharedFiles)" />
+
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="6.0.0" />
```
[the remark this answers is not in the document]
mjrousos: These are here for `ISystemWebAdapterBuilder` (which is now shared and has an IServiceCollection property). And the RemoteAppExtensions class (AddRemoteApp) is also shared now and uses AddOptions to setup RemoteAppOptions.

We probably don't need those APIs in .NET Standard (I would imagine they won't often be called from libraries), but currently all 'shared' files end up in .NET Standard, so I'd have to special case them to build for Net6 and NetFx but not NetStandard.

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/SystemWebAdapterModule.Framework.cs
```diff
@@ -2,24 +2,25 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
+using System.Collections.Generic;
 using System.Web;
+using Microsoft.Extensions.DependencyInjection;
 
 namespace Microsoft.AspNetCore.SystemWebAdapters;
 
 internal sealed class SystemWebAdapterModule : IHttpModule
```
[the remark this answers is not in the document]
mjrousos: I was actually going to do it that way originally, but it seems that the module is created multiple times for different requests which doesn't work for our scenarios since the session service needs to store state (locked session cache) between requests. I'll read up on IHttpModule lifetime and see if there's something we can do to change that behavior, but just from some ad-hoc testing I found the provider couldn't live here because it resulted in different requests having different locked session caches.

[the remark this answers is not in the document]
twsouthwick: oh good point - I don't think we can know how long the module itself is created. we probably don't wan it to dispose of anything then here and we handle that elsewhere (unless we're using a scoped instance in the module that it can own).

src/Microsoft.AspNetCore.SystemWebAdapters/Microsoft.AspNetCore.SystemWebAdapters.csproj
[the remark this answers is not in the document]
twsouthwick: Hmmm let's figure out how to fix the `.Shared` set up as I've hit that too. Maybe we drop that kind of logic and just do #if/defs in the files for the specific frameworks at the top.

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/SystemWebAdapterModule.Framework.cs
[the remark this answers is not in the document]
mjrousos: Now that you mention it, this raises some more questions about service lifetime. We're registering IHttpModules as singleton currently but based on my observation that SystemWebAdapterModule might be instantiated multiple times, we could end up calling Init multiple times on the same registered modules which probably isn't desirable. 

Because SystemWebAdapterModule is the logical place to initialize the other modules, it may make sense to register modules with scope lifetime (though services can still be singleton for sharing caches, etc.) and create a scope in SystemWebAdapterModule's init that is disposed when that module is disposed.

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/SystemWebAdaptersExtensions.Framework.cs
```diff
@@ -18,45 +20,56 @@ public static ISystemWebAdapterBuilder AddSystemWebAdapters(this HttpApplication
...
-    {
-        builder.Modules.Add(module);
         return builder;
     }
 
-    private class Builder : ISystemWebAdapterBuilder
+    internal static IServiceProvider? GetServiceProvider(this HttpApplicationState state)
     {
-        public ICollection<IHttpModule> Modules { get; } = new List<IHttpModule>();
+        var serviceProvider = state[ServicesKey] as IServiceProvider;
```
twsouthwick, line 53: how should we handle disposing the rooted container?

src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/RemoteAppAuthenticationService.cs
```diff
@@ -23,19 +23,22 @@ internal partial class RemoteAppAuthenticationService : IRemoteAppAuthentication
...
+    private readonly IOptionsSnapshot<RemoteAppAuthenticationOptions> _authOptionsSnapshot;
+    private readonly RemoteAppOptions _remoteAppOptions;
     private RemoteAppAuthenticationOptions? _options;
 
     public RemoteAppAuthenticationService(
         HttpClient client,
         IAuthenticationResultFactory resultFactory,
-        IOptionsSnapshot<RemoteAppAuthenticationOptions> options,
+        IOptionsSnapshot<RemoteAppAuthenticationOptions> authOptions,
+        IOptions<RemoteAppOptions> remoteAppOptions,
```
twsouthwick, line 34: should we be consistent here in storing the optinos itself vs the IOptions<> container? Also, why is one thing using IOptionsSnapshop<> and another IOptions<>?

src/Microsoft.AspNetCore.SystemWebAdapters/InternalSystemWebExtensions.cs
```diff
@@ -0,0 +1,95 @@
...
+using System.Diagnostics.CodeAnalysis;
+using System.Web;
+using Microsoft.AspNetCore.Http.Features;
+using Microsoft.AspNetCore.SystemWebAdapters.Internal;
+
+namespace Microsoft.AspNetCore.SystemWebAdapters;
+
+internal static class InternalSystemWebExtensions
+{
+
```
twsouthwick, line 16:
```suggestion
```

src/Microsoft.AspNetCore.SystemWebAdapters/Microsoft.AspNetCore.SystemWebAdapters.csproj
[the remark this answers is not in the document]
twsouthwick: Once this is merged, I'll take that issue and then I can remove the package dependency

twsouthwick (approved):

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/SystemWebAdaptersExtensions.Framework.cs
twsouthwick: Maybe DomainUnload in case anyone is expecting Dispose to be called to write persist any state correctly or dispose pattern isn't correctly handled.

src/Microsoft.AspNetCore.SystemWebAdapters/Authentication/RemoteAppAuthenticationService.cs
mjrousos: Auth options are a special case because they're registered by name (each auth scheme will have its own options) and we don't know which scheme to get options for until `InitializeAsync` is called. We don't currently have multiple `RemoteAppOptions` registered in DI, so I just get those options in the ctor the usual way for things we don't expect to change. That pattern doesn't work for auth options, though, since we need to get them by name.

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/SystemWebAdaptersExtensions.Framework.cs
mjrousos: Actually, looking at this a bit more I think a better option would be to use `HttpApplication`. If we make the `AddSystemWebAdapters` extension take the application rather than the application state, then we can both register the service collection in the state and use HttpApplication's `Disposed` event to clean up the provider (if one has been created) at app shutdown. This looks like a more correct way to clean up application-scoped resources.

samples/MvcApp/Global.asax.cs
```diff
@@ -16,17 +16,15 @@ protected void Application_Start()
             RouteConfig.RegisterRoutes(RouteTable.Routes);
             BundleConfig.RegisterBundles(BundleTable.Bundles);
 
-            Application.AddSystemWebAdapters()
+            this.AddSystemWebAdapters()
```
twsouthwick, line 19: I see why you want to do this, but an extension method off of `this` feels weird... we'll want to discuss that at the API review

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/SystemWebAdaptersExtensions.Framework.cs
```diff
@@ -2,61 +2,79 @@
...
+using Microsoft.AspNetCore.SystemWebAdapters;
+using Microsoft.Extensions.DependencyInjection;
 
-namespace Microsoft.AspNetCore.SystemWebAdapters;
+namespace System.Web;
 
 public static class SystemWebAdaptersExtensions
 {
-    private const string Key = "system-web-adapter-builder";
+    private const string BuilderKey = "system-web-adapter-builder";
```
twsouthwick, line 13: Do we want to enforce that the following occurs:

- [x] For any number of calls before building, the same builder is returned
- [ ] Once built, a new builder cannot be created and will fail fast
- [ ] Should someone explicitly call a `Build()` once they are done so it's not "magical"?

I think storing the builder/service provider in the same slot on the `Application` object may make that someone easier.

twsouthwick (approved):
A thought about the checks we may want to enforce, but nothing blocking

samples/MvcApp/Global.asax.cs
mjrousos: Yea, I agree. The alternative would be to just pass `this` to the method but then maybe we need to re-name the class so that it's friendlier to access the methods as statics.

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/SystemWebAdaptersExtensions.Framework.cs
mjrousos: The idea of re-using the same application state slot is interesting. If we make the user call `Build` explicitly, though, then the format of the code for setting up the system web adapter varies more between Net6 and NetFx (the build happens automatically on the Net6 side when the app's service provider is built).

Maybe we should just require a 'build', though, to make it clear.

samples/MvcApp/Global.asax.cs
mjrousos: What do you think of `SystemWebAdapterConfiguration.AddSystemWebAdapters(this)`?

PR #114, opened 2022-07-19
By default exclude all code from .NET Standard

twsouthwick:
The .NET Standard build only needs to have the references file compiled in. Everything else is infrastructure and is only needed on .NET Core/Framework/Both.

This also moves the Authentication files under Adapters to follow the convention where top level is only System.Web types, and any supporting adapters are in the Adapters folder.


src/Microsoft.AspNetCore.SystemWebAdapters/Microsoft.AspNetCore.SystemWebAdapters.csproj
```diff
@@ -23,27 +23,28 @@
...
-    <Compile Include="@(SharedFiles)" />
-
-    <!-- RemoteAppExtensions has a different namespace in net6.0 and netfx, so it can't be
-         included in .NET Standard builds. -->
-    <Compile Remove="Adapters/RemoteAppExtensions.Shared.cs" />
-    
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="6.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Options" Version="6.0.0" />
+    <Compile Include="Properties/PropertyInfo.cs" />
+    <Compile Include="Ref.Standard.cs" />
```
mjrousos, line 43: Just a thought - since the source files for the netstandard target as completely disjoint with the source for other targets, should we just have a separate csproj for the netstandard build? I don't have a strong preference either way, but it seems worth considering since the inputs are so different now.

mjrousos (approved):
LGTM. One question inline; nothing blocking.

src/Microsoft.AspNetCore.SystemWebAdapters/Microsoft.AspNetCore.SystemWebAdapters.csproj
twsouthwick: How would that reconcile the type identities?

mjrousos: We'd need to have the netstandard project/assembly have the same identity as the .NET 6 one. You'd just have one project that builds the netstandard version of the project and one that builds the net6/netfx ones. That does complicate the packaging story, though, so it's probably fine as it is.

PR #115, opened 2022-07-20
Consolidate session APIs into main library

twsouthwick:
This moves the session infrastructure into the main library. The current SessionState library existed because there was a dependency on System.Text.Json so we didn't want to force people to use that. However, now that can be opted into for a specific implementation of ISessionKeySerializer. This has been kept in the separate project which has been renamed to Microsoft.AspNetCore.SystemWebAdapters.SystemTextJson


mjrousos (approved):
LGTM

twsouthwick:
@tratcher any concerns with this change in layering?

Tratcher:
I've added the Breaking-Change label to remind us to call out the package name change in the release notes.

Tratcher:
SystemWebAdapters.SystemTextJson is an odd name since it doesn't say anything about session, serializers, etc.

twsouthwick:
Yeah, the idea I had was this would be for any JSON serialization needs we may need, not necessarily tied to session/auth/something specific.

Tratcher:
No concerns other than the naming. Cross check it with @danroth27 and @adityamandaleeka.

twsouthwick:
Closing in favor of https://github.com/dotnet/systemweb-adapters/issues/131

Issue #112, opened 2022-07-14
Feedback from initial API Review


twsouthwick:
## Summary

We had an API review and here are the list of changes we need to make:

- [x] Remove constructors from HttpContext, HttpRequest, HttpResponse https://github.com/dotnet/systemweb-adapters/pull/113
- [ ] ~~Add [Obsolete] to APIs that will cause sync over async and expose async counterparts on the ASP.NET Core build https://github.com/dotnet/systemweb-adapters/issues/164~~
- [x] Move builders to `System.Web` namespace for framework helpers and `Microsoft.AspNetCore.Builders` for the core helpers
- [x] Remove interfaces from attributes https://github.com/dotnet/systemweb-adapters/pull/113
- [x] Rename IsEnabled to IsDisabled from metadata attributes https://github.com/dotnet/systemweb-adapters/pull/113
- [x] Rename PreLoad to Preload https://github.com/dotnet/systemweb-adapters/pull/113
- [x] Make DelegatingSessionState internal (https://github.com/dotnet/systemweb-adapters/pull/134)
- [x] Make DefaultRequestHeadersToForward and DefaultResponseHeadersToForward internal (https://github.com/dotnet/systemweb-adapters/pull/120)
- [x] #138
- [x] AuthenticationEndpointPath should be a PathString on ASP.NET Core. (https://github.com/dotnet/systemweb-adapters/pull/120)
- [x] Avoid using the same name for types that are at all structurally different and will not ever have shared code calling it from top (https://github.com/dotnet/systemweb-adapters/pull/120)
- [ ] Rethink builder pattern for remote app functionality (https://github.com/dotnet/systemweb-adapters/pull/120 https://github.com/dotnet/systemweb-adapters/pull/165)

Original feedback:

- https://github.com/dotnet/aspnetcore/issues/42090
- https://github.com/dotnet/aspnetcore/issues/42091
- https://github.com/dotnet/aspnetcore/issues/42092

mjrousos:
Yea, I can make the namespace chances in #110. I'll get that updated tomorrow.

mjrousos:
For the .NET Core namespace for the extension methods, should it be `Microsoft.AspNetCore.Builders` or `Microsoft.Extensions.DependencyInjection`? The extension methods start by extending `IServiceCollection` (rather than by extending the web application builder directly).

twsouthwick:
[...]

PR #120, opened 2022-07-21
Update remote app builder pattern

twsouthwick:
**PR Title**
Update builder pattern for remote to clearly separate client and server scenarios

**PR Description**
This change makes a few API changes we've discussed around being clear around remote app options and extension methods. This introduces a new builder API (ISystemWebAdapterRemoteAppBuilder) that allows for a nested configuration pattern similar to the following:

**ASP.NET Core:**

```csharp
  .AddRemoteAppClient(remote => remote
    .Configure(options =>
    {
      options.RemoteAppUrl = ...
      options.ApiKey = ...
    }
    .AddAuthentication(true)
    .AddSession());
```

**ASP.NET Framework:**

```csharp
SystemWebAdapterConfiguration.AddSystemWebAdapters(this)
    .AddRemoteAppServer(remote => remote
        .Configure(options => options.ApiKey = ... )
        .AddAuthentication()
        .AddSession());
```

This change also separates out the options for Client/Server to be separate classes

Part of #112

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/ISystemWebAdapterRemoteAppBuilder.Shared.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/ISystemWebAdapterRemoteAppBuilder.Shared.cs
@@ -1,0 +1,11 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.AspNetCore.SystemWebAdapters;
+
+public interface ISystemWebAdapterRemoteAppBuilder
+{
+    IServiceCollection Services { get; }
+}
--- a/src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/RemoteAppServerExtensions.Framework.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/RemoteAppServerExtensions.Framework.cs
@@ -12,7 +12,7 @@ public static class RemoteAppServerExtensions
     /// Add configuration for connecting to a remote app for System.Web extensions that require a remote
     /// ASP.NET app such as remote app authentication or remote app session sharing.
     /// </summary>
-    public static ISystemWebAdapterBuilder AddRemoteAppServer(this ISystemWebAdapterBuilder builder, Action<ISystemWebAdapterRemoteAppBuilder> configure)
+    public static ISystemWebAdapterBuilder AddRemoteAppServer(this ISystemWebAdapterBuilder builder, Action<ISystemWebAdapterRemoteServerAppBuilder> configure)
     {
         if (builder is null)
         {
@@ -33,7 +33,7 @@ public static class RemoteAppServerExtensions
         return builder;
     }
 
-    public static ISystemWebAdapterRemoteAppBuilder Configure(this ISystemWebAdapterRemoteAppBuilder builder, Action<RemoteAppOptions> configure)
+    public static ISystemWebAdapterRemoteServerAppBuilder Configure(this ISystemWebAdapterRemoteServerAppBuilder builder, Action<RemoteAppOptions> configure)
     {
         if (builder is null)
         {
@@ -51,7 +51,7 @@ public static class RemoteAppServerExtensions
         return builder;
     }
 
-    private class Builder : ISystemWebAdapterRemoteAppBuilder
+    private class Builder : ISystemWebAdapterRemoteServerAppBuilder
     {
         public Builder(IServiceCollection services)
         {
--- a/src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/RemoteAppServerExtensions.Framework.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/RemoteAppServerExtensions.Framework.cs
@@ -1,18 +1,18 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using System;
 using Microsoft.AspNetCore.SystemWebAdapters;
+using Microsoft.Extensions.DependencyInjection;
 
-namespace Microsoft.Extensions.DependencyInjection;
+namespace System.Web;
 
-public static class RemoteAppExtensions
+public static class RemoteAppServerExtensions
 {
     /// <summary>
     /// Add configuration for connecting to a remote app for System.Web extensions that require a remote
     /// ASP.NET app such as remote app authentication or remote app session sharing.
     /// </summary>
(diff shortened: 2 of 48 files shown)


src/Microsoft.AspNetCore.SystemWebAdapters/Microsoft.AspNetCore.SystemWebAdapters.csproj
```diff
@@ -66,6 +66,10 @@
     <PackageReference Include="Microsoft.Extensions.Options" Version="6.0.0" />
   </ItemGroup>
 
+  <ItemGroup>
+    <SharedFiles Remove="Adapters\ISystemWebAdapterRemoteAppBuilder.Shared.cs" />
```
mjrousos, line 70: Why is this removed?

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/ISystemWebAdapterRemoteAppBuilder.Shared.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.AspNetCore.SystemWebAdapters;
+
+public interface ISystemWebAdapterRemoteAppBuilder
```
mjrousos, line 8: What if we had two remote app builder interfaces - a client version and a server version. They could have the same shape (in fact, they could both just derive from `ISystemWebAdapterRemoteAppBuilder` with no changes) but then all of the methods that `ISystemWebAdapterRemoteAppBuilder` wouldn't need client or server in their names. You could just have `AddRemoteAppAuthentication` and have the client one extend `ISystemWebAdapterRemoteAppClientBuilder` while the server one extends `ISystemWebAdapterRemoteAppServerBuilder`.

mjrousos (changes requested):
Looks good. As I looked at the changes to samples, though, it occurred to me that we might be able to simplify the API a little by having `AddRemoteAppServer` and `AddRemoteAppClient` configure different interfaces (`ISystemWebAdapterRemoteAppClientBuilder` and `ISystemWebAdapterRemoteAppServerBuilder`). Then the extension methods that work on those builders wouldn't all need the "client" and "server" designations (since they would extend different interfaces) and overall usage would simplify a little.

src/Microsoft.AspNetCore.SystemWebAdapters/Microsoft.AspNetCore.SystemWebAdapters.csproj
twsouthwick: oh VS being helpful.... I'll remove this

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/ISystemWebAdapterRemoteAppBuilder.Shared.cs
twsouthwick: those type names don't show up anywhere, though, so I'm not sure how helpful that would be. I was wondering if we even needed the "client/server" after the initial one, i.e.

**ASP.NET Core:**

```csharp
  .AddRemoteAppClient(remote => remote
    .Configure(options =>
    {
      options.RemoteAppUrl = ...
      options.ApiKey = ...
    }
    .AddAuthentication(true)
    .AddSession());
```

**ASP.NET Framework:**

```csharp
SystemWebAdapterConfiguration.AddSystemWebAdapters(this)
    .AddRemoteAppServer(remote => remote
        .Configure(options => options.ApiKey = ... )
        .AddAuthentication()
        .AddSession());
```

It's already being done as part of `AddRemoteApp[Server|Client]` so this seems like it would be fairly clear.

mjrousos: Even if the user doesn't see the type names, I think it would help because it would allow us to have the API you're describing but only the correct `Configure` or `AddAuthentication` would be usable (and suggested by Intellisense) because they would extend different types. As it is now, wouldn't the user have to choose between to `Configure` methods? That's not a great experience which is why we were going to name them Client and Server. I think if they extend different types, though, then we can get away with them both just being called 'Configure'

twsouthwick: Oh - the client stuff is only available on the core side while the framework is only available on the framework. But I see your point. I'll try that out

mjrousos: Ah, good point. I forgot that even though `ISystemWebAdapterRemoteApp` is shared, the extension methods aren't! In that case, it doesn't make as big of a difference. I still think it would be nice to have client and server versions of the builder so that it's clearer which the extension methods apply to without having to give them cumbersome names.

twsouthwick: I made a separate builder for both

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/RemoteAppServerExtensions.Framework.cs
```diff
@@ -0,0 +1,63 @@
...
+
+namespace System.Web;
+
+public static class RemoteAppServerExtensions
+{
+    /// <summary>
+    /// Add configuration for connecting to a remote app for System.Web extensions that require a remote
+    /// ASP.NET app such as remote app authentication or remote app session sharing.
+    /// </summary>
+    public static ISystemWebAdapterBuilder AddRemoteAppServer(this ISystemWebAdapterBuilder builder, Action<ISystemWebAdapterRemoteServerAppBuilder> configure)
```
Tratcher, line 15: Having both a builder callback and a builder return value on the same API is an odd mixture of patterns. Though I can also appreciate that returning ISystemWebAdapterRemoteServerAppBuilder here could have also been confusing and would break the ISystemWebAdapterBuilder chaining.

Tratcher (approved):
We'll cover this in detail in the next API review.

mjrousos (approved):
Looks good. We should update docs to reflect the changes, too. If you want to do that in a separate PR, though, this part looks good.

twsouthwick:
:) I was in the process of updating the docs so I wouldn't forget

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/RemoteAppServerExtensions.Framework.cs
twsouthwick: yeah - I had the same thought, but seemed to make more sense than having a new builder. We can discuss it at the API review

PR #121, opened 2022-07-21
Share cookies, add logout on core to sign out cookie in MVC samples

HaoK:
Note this required updating a bunch of packages to pull in the cookie/dataprotection interop 2.2 code

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -7,7 +7,6 @@
     <add key="dotnet-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" />
     <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
     <add key="myget-legacy" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/myget-legacy/nuget/v3/index.json" />
-    <add key="nuget" value="https://api.nuget.org/v3/index.json" />
   </packageSources>
   <disabledPackageSources />
 </configuration>
--- a/NuGet.config
+++ b/NuGet.config
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
     <clear />
@@ -7,6 +7,7 @@
     <add key="dotnet-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" />
     <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
     <add key="myget-legacy" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/myget-legacy/nuget/v3/index.json" />
+    <add key="nuget" value="https://api.nuget.org/v3/index.json" />
   </packageSources>
   <disabledPackageSources />
 </configuration>
--- a/samples/MvcApp/App_Start/Startup.Auth.cs
+++ b/samples/MvcApp/App_Start/Startup.Auth.cs
@@ -35,11 +35,8 @@ namespace MvcApp
                 TicketDataFormat = new AspNetTicketDataFormat(
                     new DataProtectorShim(
                         // This directory is used to share dataprotection keys between MvcApp and MvcCoreApp
-                        DataProtectionProvider.Create(new DirectoryInfo("C:\\keyDirectory"),
-                        (builder) =>
-                        {
-                            builder.SetApplicationName("CommonMvcAppName");
-                        })
+                        DataProtectionProvider.Create(new DirectoryInfo(Path.Combine(Path.GetTempPath(), "sharedkeys", "mvcapp")),
+                        builder => builder.SetApplicationName("CommonMvcAppName"))
                         .CreateProtector(
                             "Microsoft.AspNetCore.Authentication.Cookies.CookieAuthenticationMiddleware",
                             "Identity.Application",
--- a/samples/MvcApp/App_Start/Startup.Auth.cs
+++ b/samples/MvcApp/App_Start/Startup.Auth.cs
@@ -13,13 +13,6 @@ namespace MvcApp
 {
     public partial class Startup
     {
-        public static CookieAuthenticationOptions ConfigureSharedCookieAuthentication(string applicationName, CookieAuthenticationOptions cookieOptions)
-        {
-
-            return cookieOptions;
-        }
-
-
         // For more information on configuring authentication, please visit https://go.microsoft.com/fwlink/?LinkId=301864
         public void ConfigureAuth(IAppBuilder app)
         {
--- a/samples/MvcApp/App_Start/Startup.Auth.cs
+++ b/samples/MvcApp/App_Start/Startup.Auth.cs
@@ -1,8 +1,11 @@
 using System;
+using System.IO;
 using Microsoft.AspNet.Identity;
 using Microsoft.AspNet.Identity.Owin;
+using Microsoft.AspNetCore.DataProtection;
 using Microsoft.Owin;
 using Microsoft.Owin.Security.Cookies;
+using Microsoft.Owin.Security.Interop;
 using MvcApp.Models;
 using Owin;
 
@@ -24,7 +27,23 @@ namespace MvcApp
             app.UseCookieAuthentication(new CookieAuthenticationOptions
             {
                 AuthenticationType = DefaultAuthenticationTypes.ApplicationCookie,
+                CookieName = ".AspNet.ApplicationCookie",
+                CookieSameSite = SameSiteMode.Lax,
+                SlidingExpiration = true,
+                ExpireTimeSpan = TimeSpan.FromMinutes(120),
                 LoginPath = new PathString("/Account/Login"),
+                TicketDataFormat = new AspNetTicketDataFormat(
+                    new DataProtectorShim(
+                        // This directory is used to share dataprotection keys between MvcApp and MvcCoreApp
+                        DataProtectionProvider.Create(new DirectoryInfo("C:\\keyDirectory"),
+                        (builder) =>
+                        {
+                            builder.SetApplicationName("CommonMvcAppName");
+                        })
+                        .CreateProtector(
+                            "Microsoft.AspNetCore.Authentication.Cookies.CookieAuthenticationMiddleware",
+                            "Identity.Application",
+                            "v2"))),
                 Provider = new CookieAuthenticationProvider
                 {
                     // Enables the application to validate the security stamp when the user logs in.
--- a/samples/MvcApp/App_Start/Startup.Auth.cs
+++ b/samples/MvcApp/App_Start/Startup.Auth.cs
@@ -13,6 +13,13 @@ namespace MvcApp
 {
     public partial class Startup
     {
+        public static CookieAuthenticationOptions ConfigureSharedCookieAuthentication(string applicationName, CookieAuthenticationOptions cookieOptions)
+        {
+
+            return cookieOptions;
+        }
+
+
         // For more information on configuring authentication, please visit https://go.microsoft.com/fwlink/?LinkId=301864
         public void ConfigureAuth(IAppBuilder app)
         {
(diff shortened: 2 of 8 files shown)


samples/MvcCoreApp/Program.cs
```diff
@@ -1,9 +1,27 @@
...
+using Microsoft.AspNetCore.DataProtection;
+using Microsoft.AspNetCore.Identity;
 using Microsoft.AspNetCore.SystemWebAdapters;
 
 var builder = WebApplication.CreateBuilder();
 builder.Services.AddReverseProxy().LoadFromConfig(builder.Configuration.GetSection("ReverseProxy"));
 
+// These must match the data protection settings in MvcApp Startup.Auth.cs for cookie sharing to work
+builder.Services.AddDataProtection()
+    .PersistKeysToFileSystem(new DirectoryInfo("C:\\keyDirectory"))
```
HaoK, line 11: Try to figure out how to make this relative

samples/MvcApp/App_Start/Startup.Auth.cs
```diff
@@ -24,7 +27,23 @@ public void ConfigureAuth(IAppBuilder app)
             app.UseCookieAuthentication(new CookieAuthenticationOptions
             {
                 AuthenticationType = DefaultAuthenticationTypes.ApplicationCookie,
+                CookieName = ".AspNet.ApplicationCookie",
```
HaoK, line 30: Need to add docs talking about this cookie sharing in this PR

twsouthwick:
This looks to only be changes to samples, right? So it's more of a doc update on how to use the existing library?

HaoK:
Yeah I believe so, so if we configure both apps to use the same cookie, and just add a controller action to logout/clear the cookie on the Core App side, both logouts work and there's no antiforgery issue

twsouthwick:
By enabling this, it will cause all encrypted cookies to be invalid on the framework app, right? Effectively logging everyone off as a one-time part of the migration?

HaoK:
Yeah changing the cookies settings like this will cause all all existing cookies to no longer be valid

mjrousos:
I think a one-time logout is fine. It's inevitable that it will happen eventually as part of migration.

samples/MvcCoreApp/Program.cs
mjrousos: It would have to be relative to something that's available to both the ASP.NET and ASP.NET Core app, so being relative to either of them may not be very useful. I could imagine using `Path.Combine(Path.GetTempPath(), ApplicationName)` or something like that as an absolute path that both apps can find easily.

mjrousos: Also, if we add extension methods for the ASP.NET side, we could add one for ASP.NET Core, too, that takes care of setting up data protection. That might not be as important, though, since the changes here are pretty minimal and I don't think we need to put the `AddAuthentication().AddCookie` call in adapters since that's code that will live on in the ASP.NET Core app even after migration is complete.

mjrousos (changes requested):
Looks good! My only concern is that there's a lot of changes needed in the ASP.NET app; it would be great if we could encapsulate some of that in a helper method in the adapters library so that it's easier to get started with this scenario.

NuGet.config
```diff
@@ -7,6 +7,7 @@
     <add key="dotnet-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" />
     <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
     <add key="myget-legacy" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/myget-legacy/nuget/v3/index.json" />
+    <add key="nuget" value="https://api.nuget.org/v3/index.json" />
```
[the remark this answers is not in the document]
Tratcher: Thow shalt not reference nuget.org directly 😁. Ask first responders to copy the packages you need.

samples/MvcCoreApp/Program.cs
Tratcher: At least include an app subfolder like:
```suggestion
    .PersistKeysToFileSystem(new DirectoryInfo("C:\\keyDirectory\MvcApp"))
```

samples/MvcApp/packages.config
```diff
@@ -16,17 +16,48 @@
...
+  <package id="Microsoft.Extensions.Hosting.Abstractions" version="6.0.0" targetFramework="net472" />
+  <package id="Microsoft.Extensions.Logging.Abstractions" version="6.0.1" targetFramework="net472" />
+  <package id="Microsoft.Extensions.Options" version="6.0.0" targetFramework="net472" />
+  <package id="Microsoft.Extensions.Primitives" version="6.0.0" targetFramework="net472" />
   <package id="Microsoft.jQuery.Unobtrusive.Validation" version="3.2.11" targetFramework="net472" />
   <package id="Microsoft.Owin" version="4.2.1" targetFramework="net472" />
   <package id="Microsoft.Owin.Host.SystemWeb" version="4.2.1" targetFramework="net472" />
   <package id="Microsoft.Owin.Security" version="4.2.1" targetFramework="net472" />
   <package id="Microsoft.Owin.Security.Cookies" version="4.2.1" targetFramework="net472" />
+  <package id="Microsoft.Owin.Security.Interop" version="2.1.2" targetFramework="net472" />
```
Tratcher, line 42: https://www.nuget.org/packages/Microsoft.Owin.Security.Cookies.Interop/#versions-body-tab

This is flagged as deprecated and unlisted. We'll want to fix that for the latest 2.1 version.

samples/MvcApp/App_Start/Startup.Auth.cs
```diff
@@ -24,7 +27,23 @@ public void ConfigureAuth(IAppBuilder app)
...
+                        // This directory is used to share dataprotection keys between MvcApp and MvcCoreApp
+                        DataProtectionProvider.Create(new DirectoryInfo("C:\\keyDirectory"),
+                        (builder) =>
+                        {
+                            builder.SetApplicationName("CommonMvcAppName");
+                        })
+                        .CreateProtector(
+                            "Microsoft.AspNetCore.Authentication.Cookies.CookieAuthenticationMiddleware",
+                            "Identity.Application",
+                            "v2"))),
```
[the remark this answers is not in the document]
Tratcher: I think there was a change to chunking format for large cookies that's not covered here.
https://github.com/dotnet/aspnetcore/blob/de68feabb4dc84d2d5fbcfd7b3ed0b08f53216fa/src/Shared/ChunkingCookieManager/ChunkingCookieManager.cs#L42
https://github.com/aspnet/AspNetKatana/blob/e003703a8bbf003f56ad160a8d034d9560639bc8/src/Microsoft.Owin/Infrastructure/ChunkingCookieManager.cs#L46
https://github.com/dotnet/aspnetcore/blob/release/2.1/src/Security/Interop/src/ChunkingCookieManager.cs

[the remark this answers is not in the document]
Tratcher: One option would be to service Cookies.Interop with the recommended improvements. It's a stand-alone 2.1 package.

Tratcher:
Here's a fun thread for background reading: https://github.com/aspnet/AspNetKatana/issues/435

samples/MvcApp/packages.config
HaoK: Per teams chat, we aren't using the cookies interop, just the base security.interop which isn't deprecated thankfully

samples/MvcApp/App_Start/Startup.Auth.cs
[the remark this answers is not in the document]
HaoK: I think we should just go with adding whatever API we need to the adapters in the short term, can always move it somewhere later if there's demand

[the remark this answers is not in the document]
HaoK: If anything I think we should have adapters take over anything that used to be in cookie interop, since that was only built for migration scenarios anyways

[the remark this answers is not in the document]
HaoK: @mjrousos where do you think an appropriate place for the UseSharedCookie helper to live in adapters, the MVC app isn't currently using any adapter code right now right?

samples/MvcCoreApp/Program.cs
HaoK: I went with `Path.Combine(Path.GetTempPath(), "sharedkeys", "mvcapp")` which seems to work well enough

samples/MvcApp/App_Start/Startup.Auth.cs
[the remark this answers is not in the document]
HaoK: So extending IAppBuilder would require pulling in Owin for adapters I think to add the extension to IAppBuilder, is that something we are okay doing?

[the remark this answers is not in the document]
HaoK: Okay I see what I was missing, VS wasn't showing me any of the .Framework 472 files this makes more sense to me now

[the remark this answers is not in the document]
mjrousos: That's a good point about extending IAppBuilder requiring Owin. If we can come up with a nice surface area that extends `ISystemWebBuilder` that may be the better way to go. My only concern is that it may be difficult to come up with an API that has the user configure cookie sharing on the ISystemWebBuilder while configuring cookies on IAppBuilder.

An alternative would be to have a separate project (Microsoft.AspNetCore.SystemWebAdapaters.Owin or something like that) where we add IAppBuilder extensions for things like this. Then users who care about that scenario can reference that package (and they will already have OWIN dependencies) and others can ignore it.

samples/MvcApp/App_Start/Startup.Auth.cs
```diff
@@ -18,6 +21,18 @@ public void ConfigureAuth(IAppBuilder app)
...
+            var sharedDataProtectionProvider = DataProtectionProvider.Create(
+                // This directory is used to share dataprotection keys between MvcApp and MvcCoreApp
+                new DirectoryInfo(Path.Combine(Path.GetTempPath(), "sharedkeys", sharedApplicationName)),
+                    builder => builder.SetApplicationName(sharedApplicationName))
+                    .CreateProtector(
+                        "Microsoft.AspNetCore.Authentication.Cookies.CookieAuthenticationMiddleware",
+                        // Must match the Scheme name on the MvcCoreApp, i.e. IdentityConstants.ApplicationScheme
+                        "SharedCookie",
+                        "v2");
+
```
HaoK, line 35: @mjrousos I cleaned things up to remove a lot of the extraneous settings, while we could still add a ConfigureSharedCookie("appName", "sharedKeyPath", "authenticationScheme", cookieAuthenticationOptions), this at least makes the changes into 2 logical sections (configure the data protection, and then use it for the cookie ticket format.  Thoughts ?

HaoK: Perhaps its good enough with just a single helper to hide the hardcoded things out of their control with a `var sharedDataProtectionProvider = CreateSharedDataProtectionProvider(sharedKeyDirectory, appname, authScheme)`

twsouthwick: I'm liking this better - a `ConfigureSharedCookie` method seems easy to implement and would hide even more of the stuff people don't need to muck with (i.e. do they need access to the data protection provider?). Someone could always build it up themselves, but simplifying the set up as much as possible would be nice (espeically to get rid of any hard coded things)

mjrousos: I like the idea of minimizing the code the customer has to write, but unfortunately I think it will be common to need access to the data protection provider. A shared directory won't work for a lot of scenarios (anything where the ASP.NET and ASP.NET Core apps run on different machines. App Service?) so I think it's important that customers be able to provide their own data protection providers (which may use Redis or blob storage or something as a backing store).

Can we have a helper method that will make the `CreateProtector` call with all the right string constants, but allow the user to pass in a `DataProtectionProvider`?

Maybe an API like `ConfigureSharedCookie(this IAppBuilder app, CookieAuthenticationOptions options, IDataProtectionProvider dataProtectionProvider)` or something like that. It would mean pulling in a reference to OWIN stuff in the adatpers library, but we could split this API out into a separate package that users would only depend on if they wanted this sort of shared auth enabled (in which case they were using OWIN anyhow).

mjrousos (approved):
I think this is a useful update to the MVC sample. I still think it would be nice to have an extension method or two to make setting the scenario up easier for developers.

HaoK:
Are we okay with merging this as is, and I'll file a follow up issue/PR to address / add the new helpers?

samples/MvcApp/App_Start/Startup.Auth.cs
twsouthwick: See #2  - it could be put in there.

twsouthwick (approved):
I think the POC in the samples is good for now - we can add APIs in the adapters later

mjrousos:
Yep, I think we can merge this as is but let's get an issue to track adding helper methods

mjrousos:
Actually, I see you're using #121 for that. That works. Thanks!

PR #132, opened 2022-07-27
Ensure SyncRoot is not null

twsouthwick:


src/Microsoft.AspNetCore.SystemWebAdapters/Internal/NonGenericCollectionWrapper.cs
```diff
@@ -20,7 +20,7 @@ public NonGenericCollectionWrapper(ICollection<T> collection)
 
         public bool IsSynchronized => false;
 
-        public object SyncRoot => null!;
+        public object SyncRoot => _collection;
```
mjrousos, line 23: I think this is very unlikely to matter one way or another, but I looked at how we've done this in the framework before and there appears to be some precedent for just creating a new syncroot object if the wrapped collection doesn't have one: https://referencesource.microsoft.com/#mscorlib/system/collections/objectmodel/readonlydictionary.cs,434

mjrousos (commented):
This is probably fine, though just creating a new object also would work and it seems we've used that approach in the past.

What I'm more curious about is why we need this API implemented at all. How is it being used?

src/Microsoft.AspNetCore.SystemWebAdapters/Internal/NonGenericCollectionWrapper.cs
twsouthwick: I've updated to use that pattern (although we don't need it to be thread-safe)

twsouthwick:
> What I'm more curious about is why we need this API implemented at all. How is it being used?

Solely because `HttpContext.Items` is non generic and this is part of the API. If someone is using the API, it should be implemented since it's on the contract there.

mjrousos (approved):
LGTM

PR #134, opened 2022-07-29
Relayer packages to remove dependencies for adapters

### LINKED ISSUE #131: Relayer packages to minimize dependencies
## Summary

We should relayer packages to keep the core adapters (i.e. System.Web types) an isolated package, and the supporting infrastructure as a separate layer.

## Motivation and goals

The adapters need to use some external dependencies for some functionality (i.e. dependency injection, JSON serialization, etc) but that is mostly at the level of tying the services together. Specifically, this is only needed for middleware and modules that don't affect the core System.Web types that are being adapted.

In large projects (especially those with packages.config), adding dependencies low down in the stack can have undesired consequences from binding redirects to challenges with keeping versions in check (i.e. early in the process when still using packages.config). In order to help with this, we can isolate the core adapters from the infrastructure to remove any external dependencies from the adapter layer and hence the package used in supporting libraries.

RelayerPackages

## In scope

- Move all the non-System.Web types to a separate assembly. Potentially `Microsoft.AspNetCore.SystemWebAdapters.Extensions`
- Retain required interfaces (only ones identified at this point is `IBufferedResponseFeature` and `ISessionState`) in the core assembly

## Out of scope

- None

## Risks / unknowns

This will be a breaking change for how people consume the project, but will help prevent issues such as #127.

twsouthwick:
This change separates out the System.Web types into a single assembly that doesn't have any reliance on other dependencies. Any additional dependencies will be brought in by the extensions library which will only need to be added to the ASP.NET Core and ASP.NET Framework applications themselves (not any supporting libraries).

Fixes #131


twsouthwick:
@davidfowl @Tratcher @mjrousos 

I've updated the PR per our discussion this morning. There are a few types I'm including in the adapters libraries so they can potentially be used for 3rd party libraries (i.e. MessagePack support for session keys):

- ISessionState - container for session state
- ISessionSerializer - serializes ISessionState instances
- ISessionKeySerializer - serializes session key/value pairs

The following are public APIs that in the shared services. NOTE that these will be different types since I'm using shared source and consuming code couldn't build services on them as if they were the same. They do not make sense in the adapter layer:

- JsonSessionSerializerExtensions: Registers JSON support for session
- JsonSessionSerializerOptions: Options for the json serializer
- SessionSerializerOptions: General options for session serializers
- ISystemWebAdapterBuilder: Builder type for registering adapter functionality/services
- RemoteAppOptions: Options for remote app information options

I don't think we want to have same type names but not actually same types in two assemblies, so these need to be renamed or shared in a 4th package. Thoughts?

README.md
```diff
@@ -27,15 +28,13 @@ Below are the steps needed to start using the System.Web adapters with your ASP.
...
-3. If you use `HttpContext.Session`, install `Microsoft.AspNetCore.SystemWebAdapters.SessionState`
-4. In your framework application:
-   - The package installation will add a new module to your `web.config`. This module handles any customizations that are required to help migrate to .NET Core. See [this](docs/framework.md) for details on what is available here.
-5. In your class libraries:
+2. Install `Microsoft.AspNetCore.SystemWebAdapters` to supporting libraries
    - Class libraries can target .NET Standard 2.0 if desired which will ensure you are using the shared surface area
    - If you find that there's still some missing APIs, you may cross-compile with .NET Framework to maintain that behavior and handle it in .NET core in some other way
    - There should be no manual changes to enable using supported surface area of the adapters. If a member is not found, it is not currently supported on ASP.NET Core
-6. For your ASP.NET Core application:
+3. Install `Microsoft.AspNetCore.SystemWebAdapters.Extensions` to your ASP.NET Core and Framework applications
```
Tratcher, line 35: Needs updating

Tratcher (approved):

README.md
twsouthwick: I had, but apparently hadn't saved it before I pushed it up....

PR #136, opened 2022-08-02
Add async overload on .NET 6 for TransmitFile/WriteFile

twsouthwick:
This allows code to transition to a pure-async model instead of invoking sync over async.


mjrousos (approved):
Looks good other than the build errors.

twsouthwick:
I'm going to rethink this with End as well.

PR #139, opened 2022-08-08
Rework configuring timeout on remote app API

### LINKED ISSUE #133: Incorrect session endpoint when RemoteApp option has path segments
### Describe the bug
If RemoteApp url has path segments, for example https://devblogs.microsoft.com/dotnet/, the library removes the path segments when creating request for fetching session.

For example
 https://devblogs.microsoft.com/dotnet/  would create  https://devblogs.microsoft.com/fallback/adapter/session/ instead of expected https://devblogs.microsoft.com/dotnet/fallback/adapter/session/

### To Reproduce
1. Set RemoteApp to url with path segments
2. Try to fetch session

### Further technical details 
The Uri constructor discards path segments from `baseUri` argument
https://github.com/dotnet/systemweb-adapters/blob/9fc7391fdeb62cef73de29a1310013d83dc84461/src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/RemoteSession/RemoteAppSessionStateManager.cs#L38

### LINKED ISSUE #138: NetworkTimeout could probably be made an HttpClient

mjrousos:
Replaced `NetworkTimeout` with the ability to just provide an `HttpClient` as per API review. Also fixed an issue with paths on  configured base URLs getting dropped.

Fixes #133 
Fixes #138

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/RemoteSession/RemoteAppSessionStateManager.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/RemoteSession/RemoteAppSessionStateManager.cs
@@ -17,10 +17,29 @@ namespace Microsoft.AspNetCore.SystemWebAdapters.SessionState.RemoteSession;
 
 internal partial class RemoteAppSessionStateManager : ISessionManager
 {
-    private readonly HttpClient _client;
+    private readonly IHttpClientFactory _httpClientFactory;
     private readonly ISessionSerializer _serializer;
     private readonly ILogger<RemoteAppSessionStateManager> _logger;
     private readonly RemoteAppSessionStateClientOptions _options;
+    private readonly IOptions<RemoteAppOptions> _remoteAppOptions;
+    private HttpClient? _client;
+
+    private HttpClient Client
+    {
+        get
+        {
+            if (_client is null)
+            {
+                // Use the HttpClient supplied in options if one is present in options;
+                // otherwise, generate a client with an IHttpClientFactory from DI
+                _client = _remoteAppOptions.Value.BackchannelHttpClient ?? _httpClientFactory.CreateClient(SessionConstants.SessionClientName);
+                _client.BaseAddress = new Uri($"{_remoteAppOptions.Value.RemoteAppUrl.ToString().TrimEnd('/')}{_options.SessionEndpointPath}");
+                _client.DefaultRequestHeaders.Add(_remoteAppOptions.Value.ApiKeyHeader, _remoteAppOptions.Value.ApiKey);
+            }
+
+            return _client;
+        }
+    }
 
     public RemoteAppSessionStateManager(
         IHttpClientFactory httpClientFactory,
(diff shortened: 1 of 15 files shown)


src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/RemoteAppOptions.Shared.cs
```diff
@@ -37,5 +38,11 @@ public class RemoteAppOptions
     /// </summary>
     [Required]
     public Uri RemoteAppUrl { get; set; } = null!;
+
+    /// <summary>
+    /// Gets or sets an HttpClient to use for making requests to the remote app.
+    /// A new HttpClient will be automatically generated if none is provided.
```
twsouthwick, line 44:
```suggestion
    /// Gets or sets an <see cref="HttpClient"/> to use for making requests to the remote app.
    /// A new <see cref="HttpClient"/> will be automatically generated if none is provided.
```

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/Authentication/RemoteAppAuthenticationService.cs
```diff
@@ -51,7 +54,7 @@ public Task InitializeAsync(AuthenticationScheme scheme)
         // Finish initializing the http client here since the scheme won't be known
         // until the owning authentication handler is initialized.
         _options = _authOptionsSnapshot.Get(scheme.Name);
-        _client.BaseAddress = new Uri(_remoteAppOptions.RemoteAppUrl, _options.AuthenticationEndpointPath);
+        _client.BaseAddress = new Uri($"{_remoteAppOptions.RemoteAppUrl.ToString().TrimEnd('/')}{_options.AuthenticationEndpointPath}");
```
twsouthwick, line 57: should we stash this as an internal property on the options so we're not allocaitng each request with `.ToString().TrimEnd('/')`? I'm assuming this InitializeAsync is an every request thing

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/RemoteAppOptions.Shared.cs
```diff
@@ -37,5 +38,11 @@ public class RemoteAppOptions
...
+    /// </summary>
+    public HttpClient? BackchannelHttpClient { get; set; }
```
twsouthwick, line 46: Before, the timeout was on auth only. Is this something that can be shared now for all remote app stuff?

src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/RemoteSession/RemoteAppSessionStateManager.cs
```diff
@@ -64,7 +63,7 @@ public async Task<ISessionState> CreateAsync(HttpContextCore context, SessionAtt
         try
         {
             // Get or create session data
-            var response = await GetSessionDataAsync(sessionId, metadata.IsReadOnly, context, cts.Token);
+            var response = await GetSessionDataAsync(sessionId, metadata.IsReadOnly, context, context.RequestAborted);
```
twsouthwick, line 66: oh I like how this simplifies the cancellation logic - no more custom sources :)

src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/RemoteSession/RemoteAppSessionStateManager.cs
```diff
@@ -23,19 +23,21 @@ internal partial class RemoteAppSessionStateManager : ISessionManager
...
-        HttpClient client,
+        IHttpClientFactory httpClientFactory,
         ISessionSerializer serializer,
         IOptions<RemoteAppSessionStateClientOptions> sessionOptions,
         IOptions<RemoteAppOptions> remoteAppOptions,
         ILogger<RemoteAppSessionStateManager> logger)
     {
-        _client = client ?? throw new ArgumentNullException(nameof(client));
-        _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        var remoteOptions = remoteAppOptions?.Value ?? throw new ArgumentNullException(nameof(remoteAppOptions));
```
twsouthwick, line 32: do we need to get the value initially? That forces all the callbacks to be run initially at constructor injection rather than deferring to when (or if) it is actually needed

twsouthwick (approved):
LGTM! I like how it simplifies a few things. I've got a few questions but nothing blocking

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/Authentication/RemoteAppAuthenticationService.cs
mjrousos: I think it's only called once per auth handler (this method is called from the authenticaiton handler's `InitializeHandlerAsync`) but, even so, I'm fine with storing this in a field in case anything changes in the future.

mjrousos: Actually, looking more at this, it's hard to persist this anywhere because the path depends on the options retrieved in the initialize method. So, if it were called multiple times, it might have different values each time.

I'm going to leave this as-is but will test to make sure it's not called multiple times.

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/RemoteAppOptions.Shared.cs
mjrousos: Yes, I think this makes sense for all of them since it translates into an HttpClient timeout which is common to all of those scenarios.

src/Microsoft.AspNetCore.SystemWebAdapters.SessionState/RemoteSession/RemoteAppSessionStateManager.cs
mjrousos: We can wait until the HttpClient is needed. I'll update it to have a Client property which sets this up lazily.

PR #145, opened 2022-08-16
Add HttpContext.RewritePath

twsouthwick:
Part of #91


src/Microsoft.AspNetCore.SystemWebAdapters/HttpContext.cs
```diff
@@ -58,6 +58,32 @@ public IPrincipal User
...
+            path = path[..iqs];
+        }
+
+        if (!path.StartsWith("/", StringComparison.Ordinal))
+        {
+            path = "/" + path;
+        }
+
+        _context.Request.QueryString = new QueryString(qs);
+        _context.Request.Path = path;
```
Tratcher, line 84: Do we know if the input is expected to be escaped or un-escaped? Un-escaped is my guess, though it's weird that it could be un-escaped and include a query.
```suggestion
        _context.Request.Path = new PathString(path);
```

Tratcher (approved):

src/Microsoft.AspNetCore.SystemWebAdapters/HttpContext.cs
twsouthwick: Yeah, it seems weird, but it appears to be unescaped with a query. I constructed the unit tests from trying out various permutations against HttpContext

Issue #14, opened 2022-05-11
Auth: Share cookies between Framework and Core apps


mjrousos:

mjrousos:
Related abandoned draft PR: #18

HaoK:
Done as part of https://github.com/dotnet/systemweb-adapters/pull/121

HaoK:
Leaving this open to track adding a few more helpers that discussed in https://github.com/dotnet/systemweb-adapters/pull/121

adityamandaleeka:
Due to the Owin concerns, we'll skip this for RC1. Based on demand for this and other Owin-related things in the future we can consider a package like SystemWebAdapters.Owin or something.

Tratcher:
https://github.com/dotnet/AspNetCore.Docs/issues/21987#issuecomment-821132315

danroth27:
> Leaving this open to track adding a few more helpers that discussed in #121

We decided that we don't want to add new APIs for this but we do want to make sure the scenario is properly documented.

PR #157, opened 2022-08-17
Add a configure shared cookie extension method

HaoK:
Had to add a bunch of new packages to Framework

### DIFF
--- a/samples/MvcApp/App_Start/Startup.Auth.cs
+++ b/samples/MvcApp/App_Start/Startup.Auth.cs
@@ -5,6 +5,7 @@ using Microsoft.AspNet.Identity.Owin;
 using Microsoft.AspNetCore.DataProtection;
 using Microsoft.Owin;
 using Microsoft.Owin.Security.Cookies;
+using Microsoft.Owin.Security.DataHandler;
 using Microsoft.Owin.Security.Interop;
 using MvcApp.Models;
 using Owin;
@@ -13,6 +14,26 @@ namespace MvcApp
 {
     public partial class Startup
     {
+        public static CookieAuthenticationOptions ConfigureSharedCookie(string sharedApplicationName,
+            string cookieName,
+            string authScheme,
+            string sharedKeyDirectory,
+            CookieAuthenticationOptions options)
+        {
+            var sharedDataProtectionProvider = DataProtectionProvider.Create(
+                new DirectoryInfo(sharedKeyDirectory),
+                    builder => builder.SetApplicationName(sharedApplicationName))
+                    .CreateProtector(
+                        "Microsoft.AspNetCore.Authentication.Cookies.CookieAuthenticationMiddleware",
+                        authScheme,
+                        "v2");
+
+            // Settings to configure shared cookie with MvcCoreApp
+            options.CookieName = cookieName;
+            options.TicketDataFormat = new AspNetTicketDataFormat(new DataProtectorShim(sharedDataProtectionProvider));
+            return options;
+        }
+
         // For more information on configuring authentication, please visit https://go.microsoft.com/fwlink/?LinkId=301864
         public void ConfigureAuth(IAppBuilder app)
         {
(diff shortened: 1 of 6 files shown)


HaoK:
Fixes https://github.com/dotnet/systemweb-adapters/issues/14

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices.csproj
```diff
@@ -28,6 +28,10 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.DataProtection" Version="6.0.8" />
+    <PackageReference Include="Microsoft.AspNetCore.DataProtection.Extensions" Version="6.0.8" />
+    <PackageReference Include="Microsoft.Owin.Security.Cookies" Version="4.2.2" />
```
twsouthwick, line 33: This will force everyone to have the owin dependency even if they don't need it.  Should we have a package that specifically is for the Owin related helpers?

HaoK: At that point, maybe just leaving the code in the sample and adding it to the docs is a better option?

samples/MvcApp/packages.config
```diff
@@ -35,11 +35,11 @@
...
-  <package id="Microsoft.Owin" version="4.2.1" targetFramework="net472" />
-  <package id="Microsoft.Owin.Host.SystemWeb" version="4.2.1" targetFramework="net472" />
-  <package id="Microsoft.Owin.Security" version="4.2.1" targetFramework="net472" />
-  <package id="Microsoft.Owin.Security.Cookies" version="4.2.1" targetFramework="net472" />
-  <package id="Microsoft.Owin.Security.Interop" version="2.1.2" targetFramework="net472" />
+  <package id="Microsoft.Owin" version="4.2.2" targetFramework="net472" />
+  <package id="Microsoft.Owin.Host.SystemWeb" version="4.2.2" targetFramework="net472" />
+  <package id="Microsoft.Owin.Security" version="4.2.2" targetFramework="net472" />
+  <package id="Microsoft.Owin.Security.Cookies" version="4.2.2" targetFramework="net472" />
+  <package id="Microsoft.Owin.Security.Interop" version="2.2.0" targetFramework="net472" />
```
Tratcher, line 42: Stick with 2.1.x, 2.2 isn't supported.

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/Authentication/CookieExtensions.cs
```diff
@@ -0,0 +1,34 @@
+using Microsoft.AspNetCore.DataProtection;
+using Microsoft.Owin.Security.Cookies;
+using Microsoft.Owin.Security.Interop;
+using Owin;
+
+namespace Microsoft.Owin.Security.Cookies;
+
+public static class CookieExtensions
+{
+    public static CookieAuthenticationOptions ConfigureSharedCookie(this IAppBuilder app, string sharedApplicationName,
```
Tratcher, line 10: This should be an extension method on CookieAuthenticationOptions, not IAppBuilder.

samples/MvcApp/App_Start/Startup.Auth.cs
```diff
@@ -23,37 +21,30 @@ public void ConfigureAuth(IAppBuilder app)
...
                 {
-                    // Enables the application to validate the security stamp when the user logs in.
-                    // This is a security feature which is used when you change a password or add an external login to your account.  
-                    OnValidateIdentity = SecurityStampValidator.OnValidateIdentity<ApplicationUserManager, ApplicationUser>(
-                        validateInterval: TimeSpan.FromMinutes(30),
-                        regenerateIdentity: (manager, user) => user.GenerateUserIdentityAsync(manager))
-                },
-
-                // Settings to configure shared cookie with MvcCoreApp
-                CookieName = ".AspNet.ApplicationCookie",
```
Tratcher, line 53: Might as well keep setting the cookie name directly, putting it in the configure method doesn't make it easier.

HaoK:
Before I update all the packages again, @Tratcher what about @twsouthwick concern about adding OWIN dependencies to framework?  I don't think its that bad if we just leave this extension method in the samples for them to copy, its just one method.

Tratcher:
Our sample already contains everything they need. Adding the extension method there means more code they need to copy-paste. The extension method is only valuable if we ship it in box. I don't think the extra dependencies are worth it, unless we make it a separate package. That's a call to make in our next team sync.

HaoK:
Ok I'm fine if we just close this PR too, but yeah can discuss next week

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/Authentication/CookieExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+        string sharedApplicationName,
+        string authScheme,
+        string sharedKeyDirectory)
+    {
+        if (options == null)
+        {
+            throw new ArgumentNullException(nameof(options));
+        }
+
+        var sharedDataProtectionProvider = DataProtectionProvider.Create(
```
mjrousos, line 18: Rather than call `DataProtectionProvider.Create` here, it might be better to have the user provide a data protection provider since they may not want to use a shared directory for this (maybe Redis or some other solution works better in their scenario).

Though if we're removing that part of what the extension method does, maybe it does make sense to just address this with docs and samples as suggested in other comments.

twsouthwick:
@HaoK  Given https://github.com/dotnet/systemweb-adapters/issues/14#issuecomment-1226316043, should we close this for now?

PR #165, opened 2022-08-24
Separate client/server remote options

twsouthwick:
Fixes https://github.com/dotnet/systemweb-adapters/issues/112

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/RemoteAppServerOptions.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/RemoteAppServerOptions.cs
@@ -1,0 +1,25 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.ComponentModel.DataAnnotations;
+
+namespace Microsoft.AspNetCore.SystemWebAdapters;
+
+/// <summary>
+/// Options for connecting to a remote app.
+/// </summary>
+public class RemoteAppServerOptions
+{
+    /// <summary>
+    /// Gets or sets the header used to store the API key
+    /// </summary>
+    [Required]
+    public string ApiKeyHeader { get; set; } = RemoteConstants.ApiKeyHeaderName;
+
+    /// <summary>
+    /// Gets or sets an API key used to secure the endpoint
+    /// </summary>
+    [Required]
+    public string ApiKey { get; set; } = null!;
+}
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/Authentication/RemoteAppAuthenticationService.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/Authentication/RemoteAppAuthenticationService.cs
@@ -25,14 +25,15 @@ internal partial class RemoteAppAuthenticationService : IRemoteAppAuthentication
     private readonly IAuthenticationResultFactory _resultFactory;
     private readonly ILogger<RemoteAppAuthenticationService> _logger;
     private readonly IOptionsSnapshot<RemoteAppAuthenticationClientOptions> _authOptionsSnapshot;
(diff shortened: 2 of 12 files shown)


twsouthwick:
@mjrousos I believe your PR #169 will have merge conflicts here :)

CZEMacLeod:
I'm a little confused with this. I hope you don't mind me asking, but the client/server nomenclature is odd here, given that it is 2 servers, one proxied behind the other. I understand having options which only have relevant properties for each server, Framework and Core, but I would have kept a base object with the properties that are common to both - such as ApiKey and it would probably mean that #169 wouldn't have merge issues.
Perhaps RemoteAppCoreOptions and RemoteAppFrameworkOptions would make more sense, and just keep RemoteAppOptions but make it abstract?

mjrousos:
Yea, the naming here is kind of confusing. As @CZEMacLeod points out, these are both server apps. The rationale for the client/server naming here is that from the perspective of communication between the ASP.NET and ASP.NET Core apps (for proxying, remote auth, and remote session sharing), the ASP.NET Core app functions as a client sending requests to the ASP.NET app which is more like a server responding to requests.

We actually decided to not use client and server in the package names when we re-layered into Microsoft.AspNetCore.SystemWebAdapters, Microsoft.AspNetCore.SystemWebAdapters.CoreServices, and Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices to avoid this exact confusion. Based on the conversations we had on API design, though, it sounds like folks were ok with client/server names *within* those packages because the user wouldn't have any trouble knowing what to use since there'd be only one API available. I'm open to trying to standardize on Core/Framework instead of Client/Server, but honestly none of the names seem great to me. Another option would be to just drop the Client/Server distinction in names entirely now that the APIs are in separate packages, but I do think there's value in making it clear the APIs are different since they're likely to show up near one another in docs/samples/etc. despite coming from different packages.

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/RemoteAppServerOptions.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// Gets or sets the header used to store the API key
+    /// </summary>
+    [Required]
+    public string ApiKeyHeader { get; set; } = RemoteConstants.ApiKeyHeaderName;
+
+    /// <summary>
+    /// Gets or sets an API key used to secure the endpoint
+    /// </summary>
+    [Required]
+    public string ApiKey { get; set; } = null!;
```
mjrousos, line 24: Can we leave the common parts of the options in a shared library and have the client/server-specific options derive from them? Two days ago, I would have liked this approach for how it separated ASP.NET and ASP.NET Core scenarios but now that I'm adding validation logic for remote app options that will be common to both (#169), it occurs to me that there's value in having the shared portion of the options actually be shared. The two already have a shared dependency of RemoteConstants, so I don't feel too bad having them have a common base class.

RemoteAppServerOptions probably won't add anything to the base class, but that seems fine.

mjrousos (commented):
If we can leave a shared remote options type that the client/server-specific ones each derive from, it will make other ongoing work like adding option validation easier. Other than that, this looks good to me.

HaoK:
I sort of like RemoteAppFrameworkOptions vs RemoateAppCoreOptions as opposed to Client/Server personally

twsouthwick:
@HaoK let's discuss that in the API review. We had tentatively come up with client/server and this is the final change to enable that. If we want to change to framework/core it'll be easier to do after this PR across the whole code base as it'll be pretty much an exercise in find/replace.

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/RemoteAppServerOptions.cs
twsouthwick: Chatted offline about using data validations for the new validation code so it'll be an easy thing to keep separate.

mjrousos (approved):

PR #166, opened 2022-08-24
Enable logging in framework services

twsouthwick:
Now that we have services separated out, adding logging here is probably fine and consolidates some of the code pathways.


HaoK (approved):

mjrousos (approved):
LGTM

PR #169, opened 2022-08-24
Improve security of remote app connections

### LINKED ISSUE #144: Feedback from review of remote app authentication design
- [ ] Update docs to cover best practices regarding verification of the back-end (ASP.NET app). It is strongly recommended that HTTPS be used in production scenarios to verify that the identity responses are coming from a trusted source.
- [ ] Consider updating the security key used for remote app auth to have requirements like being a 128-bit hex string or being a guid or something like that so that it's harder for users to accidentally use weak keys.

mjrousos:
This helps to ensure users are choosing strong API keys by requiring them to be parsable as GUIDs. Also updates docs to explain the API key requirements and recommend HTTPS for the ASP.NET app to ensure the ASP.NET Core app can request data from it in production scenarios securely.

Fixes #144


docs/remote-authentication/remote-authentication.md
```diff
@@ -15,12 +15,13 @@ SystemWebAdapterConfiguration.AddSystemWebAdapters(this)
     .AddProxySupport(options => options.UseForwardedHeaders = true)
     .AddRemoteApp(options =>
     {
-        options.ApiKey = "MySecretKey";
+        // ApiKey is a string representing a GUID
+        options.ApiKey = "00000000-0000-0000-0000-000000000000";
```
twsouthwick, line 19: Maybe just use `Guid.NewGuid()` so that it works, but they'll need to change it for their own to actually connect. This seems like something people would copy, but not change

twsouthwick: Or, even better, why not show it being access from configuration?

docs/remote-authentication/remote-authentication.md
```diff
@@ -52,6 +55,14 @@ Finally, if the ASP.NET Core app didn't previously include authentication middle
 app.UseAuthentication();
 ```
 
+## Securing the remote app connection
```
twsouthwick, line 58: This is duplicated below it seems. Can you structure the docs so that there's a common place for that?

src/Services/RemoteAppOptions.cs
```diff
@@ -45,4 +45,9 @@ public class RemoteAppOptions
     /// </summary>
     public HttpClient? BackchannelHttpClient { get; set; }
 #endif
+
+    // To ensure secure API keys are used, enforce that
+    // keys are parsable as GUIDs
+    public static bool Validate(RemoteAppOptions options) =>
```
twsouthwick, line 51: Does this need to be public? Would internal work?

twsouthwick (approved):

docs/remote-authentication/remote-authentication.md
mjrousos: The validation will reject an empty guid, so this will fail if they copy it.

src/Services/RemoteAppOptions.cs
mjrousos: Good catch. I'll make it internal.

docs/remote-authentication/remote-authentication.md
mjrousos: Yea, I wasn't sure the best place to put this in the docs since we don't have a page just for remote app connection (that isn't session- or auth- specific). I wasn't sure if it was worth adding one just for this, but I can make one. May come in handy later anyhow.

twsouthwick: Yeah, makes sense to now have it as a separate page since we've combined the concepts. Maybe on about "Remote App Setup" or something

twsouthwick: There's the usage-guidelines docs that had some of this as well. Maybe use that as a central point?

samples/RemoteAuth/OIDC/OIDCAuth/Global.asax.cs
```diff
@@ -17,7 +17,8 @@ protected void Application_Start()
             SystemWebAdapterConfiguration.AddSystemWebAdapters(this)
                 .AddProxySupport(options => options.UseForwardedHeaders = true)
                 .AddRemoteAppServer(remote => remote
-                    .Configure(options => options.ApiKey = "test-key")
+                    // Do not re-use this ApiKey; every solution should use a unique ApiKey
```
adityamandaleeka, line 20: Suggestion: make this come from config or something to minimize copy/paste risk.

mjrousos: Done

samples/MvcCoreApp/Program.cs
```diff
@@ -21,7 +21,9 @@
         .Configure(options =>
         {
             options.RemoteAppUrl = new(builder.Configuration["ReverseProxy:Clusters:fallbackCluster:Destinations:fallbackApp:Address"]);
-            options.ApiKey = ClassLibrary.RemoteServiceUtils.ApiKey;
+
+            // Do not re-use this API key. Every solution should have a unique API key.
```
twsouthwick, line 25: do we need this here? the code itself is fine to copy

samples/MvcCoreApp/appsettings.json
```diff
@@ -1,4 +1,5 @@
 {
+  "RemoteAppApiKey": "54b69938-90dd-4f79-adcd-27fbd6f0e4b7",
```
twsouthwick, line 2: Maybe put the warning here?

twsouthwick (approved):

khellang:
Why force hexadecimal GUIDs? You could shave a lot of extra bytes on the wire by choosing more compact representations using higher-base encodings, while keeping the same amount of entropy.

If the goal is to enforce 128 bits of entropy, could you have a low-level `ReadOnlyMemory<byte>` (or just `byte[]`?) property with some convenience methods/propeties to set it using a hexadecimal GUID? Then validate the bytes instead?

mjrousos:
@khellang yea, there are a lot of options here. I prefer a string over a byte[] since that makes it easier to read from some config sources.

And we could just say "string with minimum 12 characters" or something like that, but forcing the characters to be hex digits makes it less likely users will choose a "bad" password.

I'll re-work this given @twsouthwick's recent changes so that it just validates with a regex data annotation looking for 32 hex characters (with or without dashes or parentheses).

samples/RemoteAuth/Forms/FormsAuthCore/Program.cs
```diff
@@ -11,7 +11,9 @@
         .Configure(options =>
         {
             options.RemoteAppUrl = new(builder.Configuration["ReverseProxy:Clusters:fallbackCluster:Destinations:fallbackApp:Address"]);
-            options.ApiKey = "FormsAuthSampleKey";
+
+            // Do not re-use this ApiKey; every solution should use a unique ApiKey
+            options.ApiKey = builder.Configuration["RemoteAppApiKey"];
```
twsouthwick, line 16: as above, can we put this in the config file and not here?

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/RemoteAppServerOptions.cs
```diff
@@ -11,6 +10,32 @@ namespace Microsoft.AspNetCore.SystemWebAdapters;
 /// </summary>
 public class RemoteAppServerOptions
 {
+    // Matches an empty GUID (32 0s) with dashes, parentheses, and braces optional
+    private const string EmptyGuidRegex =
```
twsouthwick, line 14: What about creating an attribute that would extend the `RegularExpressionAttribute` and could be in the shared services? Then you wouldn't have this duplication

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/RemoteAppServerOptions.cs
```diff
@@ -20,6 +45,6 @@ public class RemoteAppServerOptions
     /// <summary>
     /// Gets or sets an API key used to secure the endpoint
     /// </summary>
-    [Required]
+    [Required, RegularExpression(NonEmptyGuidRegex, ErrorMessage = "API Key must be 32 hex characters (for example a GUID)")]
```
twsouthwick, line 48: Can we check to see if it would be sufficient to just do `MinLength(16)`? That way we wouldn't be as opinionated and would probably be ensure sufficiently complex strings

twsouthwick (changes requested):

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/RemoteAppServerOptions.cs
mjrousos: That makes sense. I'll move this there.

samples/RemoteAuth/Forms/FormsAuthCore/Program.cs
mjrousos: Thanks for the reminder. I didn't have the comment there originally because I still think of comments as belonging in code rather than json, but ASP.NET Core is fine with comments in appsettings.json, so I've moved these.

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/RemoteAppServerOptions.cs
mjrousos: I can check. I'm inclined to merge this as is now and follow-up in parallel to see if it's ok to relax the restriction to just be 16 characters long so that this PR isn't blocked.

samples/RemoteAuth/Forms/FormsAuthCore/Program.cs
twsouthwick: > ASP.NET Core is fine with comments in appsettings.json

Good to know! I wasn't sure if it would yell at you if you put comments there

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/RemoteAppServerOptions.cs
twsouthwick: Sounds good. This requirement will be a subset of a `MinLength(16)`, so would be fine to change to that later.

samples/MvcCoreApp/Program.cs
mjrousos: Moved to config

twsouthwick (approved):

Issue #156, opened 2022-08-17
Host header is changed too late when using UseHostHeaderForRequestUrl


khellang:
### Describe the bug

When running the backend ASP.NET app with the following app setting:

```csharp
<add key="aspnet:UseHostHeaderForRequestUrl" value="true" />
```

The `Host` header from the backend app will be used to build `Request.Url` instead of the forwarded proxy host.

This seems to be because the `Host` header from `X-Forwarded-Host` is set too late in the pipeline and that `Request.Url` [has already been calculated](https://referencesource.microsoft.com/#System.Web/HttpRequest.cs,98945de8f620647a).

### To Reproduce

Add the following app setting to the backend ASP.NET app:

```csharp
<add key="aspnet:UseHostHeaderForRequestUrl" value="true" />
```

Inspect `Request.Url` in the app. It should show the host of the backend app, not the proxy.

### Further technical details 

*ASP.NET Framework Application*:
- Technologies and versions used (i.e. MVC/WebForms/etc): MVC, WebForms, WebAPI
- .NET Framework Version: 4.8.4515.0
- IIS Version: 10.0.19041.320
- Windows Version: 21H2

*ASP.NET Core Application*:
- Targeted .NET version: net6.0
- .NET SDK version: 6.0.400

khellang:
I tried setting a breakpoint where the `Url` ends up being built, this is the stack trace I'm getting:

```
System.Web.dll!System.Web.HttpRequest.Url.get() Line 923
	at System.Web\HttpRequest.cs(923)
Microsoft.Owin.Host.SystemWeb.dll!Microsoft.Owin.Host.SystemWeb.OwinCallContext.Microsoft.Owin.Host.SystemWeb.CallEnvironment.AspNetDictionary.IPropertySource.GetRequestQueryString() Line 96
	at /_/src/Microsoft.Owin.Host.SystemWeb/OwinCallContext.Environment.cs(96)
Microsoft.Owin.Host.SystemWeb.dll!Microsoft.Owin.Host.SystemWeb.CallEnvironment.AspNetDictionary.RequestQueryString.get() Line 278
	at /_/src/Microsoft.Owin.Host.SystemWeb/CallEnvironment/AspNetDictionary.Generated.cs(278)
Microsoft.Owin.Host.SystemWeb.dll!Microsoft.Owin.Host.SystemWeb.CallEnvironment.AspNetDictionary.PropertiesTryGetValue(string key, out object value) Line 1078
[...]

PR #171, opened 2022-08-25
Use reflection to set host header when forwarding

HaoK:
Fixes https://github.com/dotnet/systemweb-adapters/issues/156

Verified in debugger that the reflection codepath is not hit on normal requests to framework app, but is hit and updates the url correctly in the coreapp:

image

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
@@ -27,9 +27,7 @@ internal class ProxyHeaderModule : IHttpModule
 
     private readonly IOptions<ProxyOptions> _options;
 
-    private FieldInfo _wrField;
-    private Type _iis7WorkerRequestType;
-    private FieldInfo _knownRequestHeadersField;
+    private ReflectionHelper? _reflectionHelper;
 
     public ProxyHeaderModule(IOptions<ProxyOptions> options)
     {
@@ -46,9 +44,7 @@ internal class ProxyHeaderModule : IHttpModule
 
         if (options.UseForwardedHeaders)
         {
-            _wrField = typeof(HttpRequest).GetField("_wr", BindingFlags.NonPublic | BindingFlags.Instance);
-            _iis7WorkerRequestType = Assembly.GetAssembly(typeof(HttpRequest)).GetType("System.Web.Hosting.IIS7WorkerRequest");
-            _knownRequestHeadersField = _iis7WorkerRequestType.GetField("_knownRequestHeaders", BindingFlags.NonPublic | BindingFlags.Instance);
+            _reflectionHelper = new ReflectionHelper();
 
             context.BeginRequest += (s, e) =>
             {
@@ -89,7 +85,7 @@ internal class ProxyHeaderModule : IHttpModule
 
             requestHeaders[Host] = host;
 
-            SetHostHeader(request, host);
+            _reflectionHelper?.SetHostHeader(request, host);
         }
     }
 
@@ -102,21 +98,8 @@ internal class ProxyHeaderModule : IHttpModule
         serverVariables.Set(ServerHttps, values.Https);
         requestHeaders[Host] = values.Host;
 
-        SetHostHeader(request, values.Host);
+        _reflectionHelper?.SetHostHeader(request, values.Host);
     }
-
-    private void SetHostHeader(HttpRequest? request, string host)
-    {
-        // Need to use reflection to force HttpRequest to update the known headers
-        if (request != null)
-        {
-            var workerRequest = _wrField.GetValue(request);
-            var knownRequestHeaders = (string[])_knownRequestHeadersField.GetValue(workerRequest);
-            // This is the Host index
-            knownRequestHeaders[HttpWorkerRequest.HeaderHost] = host;
-        }
-    }
-
 
     private static void UseForwardedFor(NameValueCollection requestHeaders, NameValueCollection serverVariables)
     {
@@ -124,6 +107,31 @@ internal class ProxyHeaderModule : IHttpModule
         {
             serverVariables.Set("REMOTE_ADDR", remote);
             serverVariables.Set("REMOTE_HOST", remote);
+        }
+    }
+
+    private class ReflectionHelper
+    {
+        private readonly FieldInfo _wrField;
+        private readonly Type _iis7WorkerRequestType;
+        private readonly FieldInfo _knownRequestHeadersField;
+
+        public ReflectionHelper()
+        {
+            _wrField = typeof(HttpRequest).GetField("_wr", BindingFlags.NonPublic | BindingFlags.Instance);
+            _iis7WorkerRequestType = Assembly.GetAssembly(typeof(HttpRequest)).GetType("System.Web.Hosting.IIS7WorkerRequest");
+            _knownRequestHeadersField = _iis7WorkerRequestType.GetField("_knownRequestHeaders", BindingFlags.NonPublic | BindingFlags.Instance);
+        }
+        public void SetHostHeader(HttpRequest? request, string host)
+        {
+            // Need to use reflection to force HttpRequest to update the known headers
+            if (request != null)
+            {
+                var workerRequest = _wrField.GetValue(request);
+                var knownRequestHeaders = (string[])_knownRequestHeadersField.GetValue(workerRequest);
+                // This is the Host index
+                knownRequestHeaders[HttpWorkerRequest.HeaderHost] = host;
+            }
         }
     }
 
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
@@ -117,14 +117,13 @@ internal class ProxyHeaderModule : IHttpModule
     private class ReflectionHelper
     {
         private readonly FieldInfo _wrField;
-        private readonly Type _iis7WorkerRequestType;
         private readonly FieldInfo _knownRequestHeadersField;
 
         public ReflectionHelper()
         {
             _wrField = typeof(HttpRequest).GetField("_wr", BindingFlags.NonPublic | BindingFlags.Instance);
-            _iis7WorkerRequestType = Assembly.GetAssembly(typeof(HttpRequest)).GetType("System.Web.Hosting.IIS7WorkerRequest");
-            _knownRequestHeadersField = _iis7WorkerRequestType.GetField("_knownRequestHeaders", BindingFlags.NonPublic | BindingFlags.Instance);
+            var type = Assembly.GetAssembly(typeof(HttpRequest)).GetType("System.Web.Hosting.IIS7WorkerRequest");
+            _knownRequestHeadersField = type.GetField("_knownRequestHeaders", BindingFlags.NonPublic | BindingFlags.Instance);
         }
         
         public void SetHostHeader(HttpRequest request, string host)
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
@@ -1,9 +1,9 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using System;
 using System.Collections.Specialized;
 using System.Globalization;
+using System.Reflection;
 using System.Web;
 using Microsoft.Extensions.Options;
 
@@ -27,6 +27,10 @@ internal class ProxyHeaderModule : IHttpModule
 
     private readonly IOptions<ProxyOptions> _options;
 
+    private FieldInfo _wrField;
+    private Type _iis7WorkerRequestType;
+    private FieldInfo _knownRequestHeadersField;
+
     public ProxyHeaderModule(IOptions<ProxyOptions> options)
     {
         _options = options ?? throw new ArgumentNullException(nameof(options));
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
```diff
@@ -80,6 +89,15 @@ public void UseHeaders(NameValueCollection requestHeaders, NameValueCollection s
             serverVariables.Set(ServerHttps, value.IsSecure ? On : Off);
 
             requestHeaders[Host] = host;
+
+            // Need to use reflection to force HttpRequest to update the known headers
+            if (request != null)
```
twsouthwick, line 94: This needs to be done on the code path for line 111

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
```diff
@@ -80,6 +89,15 @@ public void UseHeaders(NameValueCollection requestHeaders, NameValueCollection s
...
+            {
+                var workerRequest = _wrField.GetValue(request);
```
twsouthwick, line 96: Could this be it's own method of along the lines of "SetHostHeader"?

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
```diff
@@ -42,10 +46,15 @@ public void Init(HttpApplication context)
 
         if (options.UseForwardedHeaders)
         {
+            _wrField = typeof(HttpRequest).GetField("_wr", BindingFlags.NonPublic | BindingFlags.Instance);
```
twsouthwick, line 49: I'd prefer some sort of helper class/struct that would handle all the reflection stuff behind the scenes so the module doesn't have to feel so "hacky"

twsouthwick (changes requested):

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
HaoK: Yeah makes sense now that we call this in two places, done

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
HaoK: Thanks for catching this

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
HaoK: Oh, are you talking about initializing the fields/type I guess?

HaoK: I mean I can move this into some kind of InitializeReflection if you really want

twsouthwick:
fyi - you're previous build failed in part due to using outdated vms. I have a PR https://github.com/dotnet/systemweb-adapters/pull/172 to fix that.

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
HaoK: So if I hear you, you prefer some kind of new HostReflectionHelper class here that we can pass in the request

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
```diff
@@ -55,12 +63,12 @@ public void Init(HttpApplication context)
...
             {
                 var request = ((HttpApplication)s).Context.Request;
-                UseOptions(values, request.Headers, request.ServerVariables);
+                UseOptions(values, request.Headers, request.ServerVariables, request);
             };
         }
     }
 
-    public void UseHeaders(NameValueCollection requestHeaders, NameValueCollection serverVariables)
+    public void UseHeaders(NameValueCollection requestHeaders, NameValueCollection serverVariables, HttpRequest? request = null)
```
twsouthwick, line 71: this class is tested with the public methods (not with Init as HttpApplication cannot be mocked)

twsouthwick: so tests will probably fail :)

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
twsouthwick: I'd prefer the helper class direction

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
HaoK: So I wasn't planning on adding unit tests for this, once we have playwright setup I could verify this behavior via the contact page showing the correct request.url for the coreapp and the app.

HaoK: The tests use null request which just no-op the reflection

twsouthwick: I was referring to the existing tests. Do they still pass? This method was designed to be fully testable (hence, passing the headers/server variables rather than the request).

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
```diff
@@ -42,10 +44,12 @@ public void Init(HttpApplication context)
 
         if (options.UseForwardedHeaders)
         {
+            _reflectionHelper = new ReflectionHelper();
```
twsouthwick, line 47: This will only allow the reflection helper stuff to run when in this code path, but it should be available for both. Can't this be initialized in the constructor?

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
```diff
@@ -103,6 +110,31 @@ private static void UseForwardedFor(NameValueCollection requestHeaders, NameValu
...
+        private readonly Type _iis7WorkerRequestType;
+        private readonly FieldInfo _knownRequestHeadersField;
+
+        public ReflectionHelper()
+        {
+            _wrField = typeof(HttpRequest).GetField("_wr", BindingFlags.NonPublic | BindingFlags.Instance);
+            _iis7WorkerRequestType = Assembly.GetAssembly(typeof(HttpRequest)).GetType("System.Web.Hosting.IIS7WorkerRequest");
+            _knownRequestHeadersField = _iis7WorkerRequestType.GetField("_knownRequestHeaders", BindingFlags.NonPublic | BindingFlags.Instance);
+        }
+        public void SetHostHeader(HttpRequest? request, string host)
```
twsouthwick, line 125: Can we require this to be non-nullable?

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
```diff
@@ -103,6 +110,31 @@ private static void UseForwardedFor(NameValueCollection requestHeaders, NameValu
...
+        {
+            // Need to use reflection to force HttpRequest to update the known headers
+            if (request != null)
+            {
+                var workerRequest = _wrField.GetValue(request);
+                var knownRequestHeaders = (string[])_knownRequestHeadersField.GetValue(workerRequest);
```
twsouthwick, line 131: is there any chance this will be null?

khellang: In theory, yes. It's `null` until the headers are accessed the first time. This normally happens *very* early in the pipeline, during request validation, but I think there might be a way to turn off request validation, even if it's super frowned upon 😅

twsouthwick: sounds like we should probably check for null before updating it. If it's null, then it should propagate fine and this workaround wouldn't be needed, right?

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
twsouthwick: does it need to be called there then? especially since both pathways need it, maybe call it after returning and still in the BetginRequest callback,

ie

```csharp
            context.BeginRequest += (s, e) =>
            {
                var request = ((HttpApplication)s).Context.Request;
                UseOptions(values, request.Headers, request.ServerVariables);
                reflectionHelper.SetHostName(request.ServerVariables[Host]);
            };
```

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
khellang: Yeah, if it's null, it can just return early as the correct value will be read the first time 👍

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
HaoK: Yeah the existing methods pass fine

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
HaoK: Do we need to run the reflection path when we aren't forwarding headers?  At least from my reading of the code, we don't do need to do anything expect when the host is forwarded, since that's all the reflection helper is setting anywyas

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
HaoK: Its the unit tests that pass in null, I was being lazy and not wanting to change the tests to have to mock some kind of httprequest

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
HaoK: Yeah if its null we don't need to run this at all, so good idea to just do nothing if we don't get the array

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
HaoK: Okay I will just move this to the constructor

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
HaoK: I made this non nullable and just had the callers null check for the unit tests

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
```diff
@@ -103,6 +114,31 @@ private static void UseForwardedFor(NameValueCollection requestHeaders, NameValu
...
+        private readonly FieldInfo _wrField;
+        private readonly Type _iis7WorkerRequestType;
+        private readonly FieldInfo _knownRequestHeadersField;
+
+        public ReflectionHelper()
+        {
+            _wrField = typeof(HttpRequest).GetField("_wr", BindingFlags.NonPublic | BindingFlags.Instance);
+            _iis7WorkerRequestType = Assembly.GetAssembly(typeof(HttpRequest)).GetType("System.Web.Hosting.IIS7WorkerRequest");
+            _knownRequestHeadersField = _iis7WorkerRequestType.GetField("_knownRequestHeaders", BindingFlags.NonPublic | BindingFlags.Instance);
+        }
```
Tratcher, line 128:
```suggestion
        }
        
```

Tratcher (approved):

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
```diff
@@ -103,6 +114,31 @@ private static void UseForwardedFor(NameValueCollection requestHeaders, NameValu
         }
     }
 
+    private class ReflectionHelper
+    {
+        private readonly FieldInfo _wrField;
+        private readonly Type _iis7WorkerRequestType;
```
twsouthwick, line 120: nit: is this actually needed? isn't it just to find the second field?

twsouthwick (approved):

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/ProxyHeaderModule.cs
HaoK: Good point, will remove

PR #177, opened 2022-08-28
Use a common remote module to ensure standard checks

twsouthwick:
This change creates a module base class that enables exposing endpoints from the service in a standard way. This pattern is testable as well as ensures things like API key checks are done as expected.


mjrousos (approved):
LGTM

PR #178, opened 2022-08-28
Use a shared client for remote app connections

twsouthwick:
This change brings the configuration of the HttpClient into a shared
location for all the remote app connections. Now, any system that needs
to connect to the remote app can do so by just retrieving the client via
the IHttpClientFactory by name and the path/URI/backchannel will be
configured for them.

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/RemoteAppClientOptions.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/RemoteAppClientOptions.cs
@@ -31,8 +31,7 @@ public class RemoteAppClientOptions
     public Uri RemoteAppUrl { get; set; } = null!;
 
     /// <summary>
-    /// Gets or sets an <see cref="HttpClient"/> to use for making requests to the remote app.
-    /// A new <see cref="HttpClient"/> will be automatically generated if none is provided.
+    /// Gets or sets an <see cref="HttpMessageHandler"/> to use for making requests to the remote app.
     /// </summary>
-    public HttpClient? BackchannelHttpClient { get; set; }
+    public HttpMessageHandler? BackchannelHandler { get; set; }
 }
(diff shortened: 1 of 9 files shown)


src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/RemoteAppClientOptions.cs
```diff
@@ -31,8 +31,7 @@ public class RemoteAppClientOptions
     public Uri RemoteAppUrl { get; set; } = null!;
 
     /// <summary>
-    /// Gets or sets an <see cref="HttpClient"/> to use for making requests to the remote app.
-    /// A new <see cref="HttpClient"/> will be automatically generated if none is provided.
+    /// Gets or sets an <see cref="HttpMessageHandler"/> to use for making requests to the remote app.
     /// </summary>
-    public HttpClient? BackchannelHttpClient { get; set; }
+    public HttpMessageHandler? BackchannelHandler { get; set; }
```
twsouthwick, line 36: @mjrousos I had to change this to a handler so that it could be fit into the HttpClientFactory pipeline. It seems reasonable to me, but maybe I'm missing something?

mjrousos: There was some discussion about whether to use an HttpClient or HttpMessageHandler during the API review. We landed on an HttpClient since that provides the most flexibility ([link](https://github.com/dotnet/aspnetcore/issues/42092#issuecomment-1185965114)) but I don't think there's a large difference here. I'm ok with this change unless @halter73 has push back.

twsouthwick: I'm including it on the API review

mjrousos (approved):
LGTM

Issue #83, opened 2022-06-30
Add AspNet Core side shim for System.Web.VirtualPathUtility


CZEMacLeod:
## Summary

Add the `System.Web.VirtualPathUtility` class to systemweb adapters to allow the use from AspNet Core and class libraries whilst migrating an application.

## Motivation and goals

Within our application we make use of this class, mainly to handle dynamic paths and in areas where we need to return absolute paths regardless of the request url (e.g. in webapi call, middleware, partial views etc.).

## In scope

Handle the following methods on `System.Web.VirtualPathUtility`
```cs
namespace System.Web;

public static class VirtualPathUtility
{
  public static string Combine(string basePath, string relativePath);
  public static string GetDirectory(string virtualPath);
  public static string GetExtension(string virtualPath);
  public static string GetFileName(string virtualPath);
  public static bool IsAppRelative(string virtualPath);
  public static string MakeRelative(string fromPath, string toPath);
  public static string ToAbsolute(string virtualPath);
  public static string ToAppRelative(string virtualPath);
}
```

## Out of scope

The overloads which allow you to include the `applicationPath` would not be needed (at least not for my project).

## Risks / unknowns

Most of these functions relate to translating paths starting ~/ to the base url which should be available with #81 as HttpRuntime.AppDomainAppVirtualPath.
The naïve implementation could use UrlBuilder and/or combinations of StartsWith and SubString to implement the main functions.
The Get* methods would probably need to get the IFileProvider interface registered to get the appropriate physical file paths and names for virtual paths.
GetFileName and GetExtension should work regardless of whether the file resolved to a physical file or not.
Most of these functions are no longer in use in web applications since it is now common to have each web application at the root, rather than installed as a sub site in IIS so this is largely for legacy projects where these abstractions have not yet been removed, or where configuration files or similar use 'app relative paths' like ~/App_Data/myconfig.json

## Examples

```cs
public static class ThemeClaims
{
	public static string GetThemeOrDefault(this ClaimsIdentity identity, string @default = "default")
	{
		var claim = identity?.FindFirst("myapp:Theme");
		if (claim is null) return @default;
		return claim.Value;
	}
}

public class ThemeModel
{
	public string Theme { get; set; } = 
		(System.Web.HttpContext.Current.User?.Identity as ClaimsIdentity).GetThemeOrDefault();
	public string ThemeURL { 
		get => System.Web.VirtualPathUtility
[...]

PR #179, opened 2022-08-28
Add System.Web.VirtualPathUtility

CZEMacLeod:
Add System.Web.VirtualPathUtility implementation for AspNet Core
Add an implementation of System.Web.VirtualPathUtility based on the codebase for net framework

Adds System.Web.VirtualPathUtility based (in part) on
System.Web.Util.UrlPath code at https://github.com/microsoft/referencesource/blob/master/System.Web/Util/UrlPath.cs
System.Web.Util.StringUtil at https://github.com/microsoft/referencesource/blob/master/System.Web/Util/StringUtil.cs

Addresses #83


src/Microsoft.AspNetCore.SystemWebAdapters/Ref.Standard.cs
```diff
@@ -448,6 +448,106 @@ public enum SameSiteMode
         None = 0,
         Strict = 2,
     }
+
+    /// <summary>
+    /// Provides utility methods for common virtual path operations.
+    /// </summary>
+    public static class VirtualPathUtility
```
twsouthwick, line 455: Can you run the `./update-apis.ps1` script? This file is autogenerated (along with the type forwards)

src/Microsoft.AspNetCore.SystemWebAdapters/VirtualPathUtility.cs
```diff
@@ -0,0 +1,702 @@
...
+/// <summary>
+/// Provides utility methods for common virtual path operations.
+/// </summary>
+/// <remarks>
+/// Portions of this code are based on code originally Copyright (c) 1999 Microsoft Corporation
+/// System.Web.Util.UrlPath code at https://github.com/microsoft/referencesource/blob/master/System.Web/Util/UrlPath.cs
+/// System.Web.Util.StringUtil at https://github.com/microsoft/referencesource/blob/master/System.Web/Util/StringUtil.cs
+/// These files are released under an MIT licence according to https://github.com/microsoft/referencesource#license
+/// </remarks>
+public static class VirtualPathUtility
```
twsouthwick, line 18: Can you add some tests for this?

src/Microsoft.AspNetCore.SystemWebAdapters/Ref.Standard.cs
CZEMacLeod: I did run that, but it seemed to overwrite other customizations to those files, so I just made the changes manually.
I didn't want to mix up changes for VPU with other stuff.
TypeForwards.Framework:

- Loses the comments and licence info
- Adds type forwards for Microsoft.AspNetCore.SystemWebAdapters.ISessionKeySerializer, Microsoft.AspNetCore.SystemWebAdapters.ISessionState and Microsoft.AspNetCore.SystemWebAdapters.SessionState.Serialization.ISessionSerializer
- Loses the #if NET472_OR_GREATER for System.Web.SameSiteMode

Ref.Standard:

- Adds implicit operators to HttpRequestBase for Microsoft.AspNetCore.Http.HttpRequest and HttpResponseBase for Microsoft.AspNetCore.Http.HttpResponse

Also, I would point out that the error message is slightly oddly worded: "Only _support_ when running on ASP.NET Core or System.Web" which is why I used a const with "Only _supported_ when running on ASP.NET Core or System.Web"

I noticed that the netstandard/ref version doesn't include the doc comments, so you don't get intellisense in class libraries targeting netstandard2.0 hence why I kept the comments in too.

I think the only other difference between my manual changes and the script is making the ref class partial, but I don't see that actually being used for anything.

Let me know how you want me to proceed.

src/Microsoft.AspNetCore.SystemWebAdapters/VirtualPathUtility.cs
CZEMacLeod: I'll see what I can come up with. There aren't really any good examples of what should become what for these calls that I could find. I did add some exercising code to RequestInfo in the class library example to check it working from both sides and I did some checks of my own to my implementation, but since it relies on HttpRuntime.AppDomainAppVirtualPath for some of its calls, it is harder to test all scenarios.

CZEMacLeod: I've added some tests which I think covers most items and scenarios - and indeed caught a couple of things that didn't quite work the same as net4x, which are now fixed.

src/Microsoft.AspNetCore.SystemWebAdapters/Ref.Standard.cs
twsouthwick: Good catch - I'd been the main person using it so I got used to the workarounds. I've added #181 to fix it up so it should work as expected (and the build will run it to to ensure we didn't miss anything).

> Also, I would point out that the error message is slightly oddly worded:
Yeah, I kept forgetting to fix that. Thanks for point it out - it's fixed in #181

> I noticed that the netstandard/ref version doesn't include the doc comments
I hadn't realized that! Thanks. I'd like to figure out a fix that brings that over for all of them. I'll create a separate issue to track that

> Let me know how you want me to proceed.
For now, it's fine, but once #181 is merged, please merge it in and run the script so things will stay in sync

src/Microsoft.AspNetCore.SystemWebAdapters/Microsoft.AspNetCore.SystemWebAdapters.csproj
```diff
@@ -28,20 +28,33 @@
...
     <Compile Include="Ref.Standard.cs" />
+    <EmbeddedResource Remove="**/*" />
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'net6.0' ">
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
 
     <PackageReference Include="System.Runtime.Caching" Version="6.0.0" />
 
+    <EmbeddedResource Update="VirtualPathUtilityErrors.resx">
```
twsouthwick, line 39: @Tratcher I'd been told early on in the project just to inline error strings. Should we continue that or should we move things to a resource?

test/Microsoft.AspNetCore.SystemWebAdapters.Tests/VirtualPathUtilityTests.cs
```diff
@@ -0,0 +1,212 @@
...
+        {
+            Assert.Equal(expected, (string?)VirtualPathUtility.ToAbsolute(virtualPath, "/"));
+        }
+
+        [Fact]
+        public void ToAbsoluteError()
+        {
+#if NET40_OR_GREATER
+			// This does not match the documentation
+			Assert.Throws<ArgumentException>(() => VirtualPathUtility.ToAbsolute("hello", "/"));
```
twsouthwick, line 22: Is there a way to make the exceptions match?

test/Microsoft.AspNetCore.SystemWebAdapters.Tests/VirtualPathUtilityTests.cs
```diff
@@ -0,0 +1,212 @@
...
+        [Theory]
+        public void ToAbsolute(string virtualPath, string expected)
+        {
+            Assert.Equal(expected, (string?)VirtualPathUtility.ToAbsolute(virtualPath, "/"));
+        }
+
+        [Fact]
+        public void ToAbsoluteError()
+        {
+#if NET40_OR_GREATER
```
twsouthwick, line 20: I've generally not been running the tests on framework mostly because the mocking ends up being drastically different. 

These tests could potentially be more exhaustive by changing the implementation to be an (internal) instance class that takes in a virtual directory class and the static calls just go through a cached instance that has the result of HttpRuntime.AppVirtualDirectory

test/Microsoft.AspNetCore.SystemWebAdapters.Tests/VirtualPathUtilityTests.cs
```diff
@@ -0,0 +1,212 @@
...
+        [InlineData("/", "/")]
+        [InlineData("/test/", "/test")]
+        [Theory]
+        public void RemoveTrailingSlash(string virtualPath, string expected)
+        {
+            Assert.Equal(expected, (string?)VirtualPathUtility.RemoveTrailingSlash(virtualPath));
+        }
+
+#if NET6_0_OR_GREATER
+        [InlineData("~", "/")]
```
twsouthwick, line 76: Why are these conditioned?

twsouthwick (commented):
This looks really good! It's reassuring seeing those tests that behavior is aligned :)

src/Microsoft.AspNetCore.SystemWebAdapters/Microsoft.AspNetCore.SystemWebAdapters.csproj
CZEMacLeod: I'm happy to inline them, perhaps with private constants for the ones that are shared. When I first started building this, I took across all the code from the referenced private classes and that used a lot more error messages - but once it was stripped down to just the stuff required, there weren't actually very many left.

Tratcher: Yes, inline them. resx files are more trouble than they're worth at this stage.

test/Microsoft.AspNetCore.SystemWebAdapters.Tests/VirtualPathUtilityTests.cs
CZEMacLeod: I can make the exceptions match - just net4x doesn't throw the exception that the docs site says it does and I went by the docs in this case.
https://docs.microsoft.com/en-us/dotnet/api/system.web.virtualpathutility.toabsolute?view=netframework-4.8 says it throws an `ArgumentOutOfRangeException` but the framework actually throws an `ArgumentException` (or sometimes an `ArgumentNullException`)

test/Microsoft.AspNetCore.SystemWebAdapters.Tests/VirtualPathUtilityTests.cs
CZEMacLeod: I was running the same tests on a net48 project to ensure I was getting the same results.
The extra data is conditioned because under net48, these calls fail due to the fact that ~ is not defined unless you are running under a web application context (which the tests weren't) so they were failing on net48. I setup a webapi page that ran through the same data and showed the outputs - this is actually how I checked the exceptions and values matched.

twsouthwick: Ah - that makes sense part of why I haven't tended to run tests on framework. I see value in it to make sure we don't regress, so if you'd like to keep that running on .net 48 that's fine. Just add a comment as to why they're conditioned

test/Microsoft.AspNetCore.SystemWebAdapters.Tests/VirtualPathUtilityTests.cs
CZEMacLeod: I'm happy to take the conditionals out, but they were useful when I was running to compare a net48 project against the actual System.Web implementation.

test/Microsoft.AspNetCore.SystemWebAdapters.Tests/VirtualPathUtilityTests.cs
CZEMacLeod: I've updated the code to match the same exceptions now. I think I also caught one more scenario which wasn't being caught while I was at it and removed the resource file.

src/Microsoft.AspNetCore.SystemWebAdapters/Microsoft.AspNetCore.SystemWebAdapters.csproj
CZEMacLeod: I've removed the resource file

src/Microsoft.AspNetCore.SystemWebAdapters/Ref.Standard.cs
twsouthwick: #181 is merged - please merge it in and run the script

test/Microsoft.AspNetCore.SystemWebAdapters.Tests/VirtualPathUtilityTests.cs
```diff
@@ -0,0 +1,215 @@
...
+        public void GetDirectory(string virtualPath, string expected)
+        {
+            Assert.Equal(expected, (string?)VirtualPathUtility.GetDirectory(virtualPath));
+        }
+
+        [Fact]
+        public void GetDirectoryError()
+        {
+			Assert.Throws<ArgumentNullException>(() => VirtualPathUtility.GetDirectory(null!));
+			Assert.Throws<ArgumentNullException>(() => VirtualPathUtility.GetDirectory(""));
```
twsouthwick, line 97:
```suggestion
	Assert.Throws<ArgumentNullException>(() => VirtualPathUtility.GetDirectory(null!));
	Assert.Throws<ArgumentNullException>(() => VirtualPathUtility.GetDirectory(""));
```

test/Microsoft.AspNetCore.SystemWebAdapters.Tests/VirtualPathUtilityTests.cs
```diff
@@ -0,0 +1,215 @@
...
+        }
+
+        [Fact]
+        public void ToAbsoluteError()
+        {
+            // This does not match the documentation
+            // https://docs.microsoft.com/en-us/dotnet/api/system.web.virtualpathutility.toabsolute?view=netframework-4.8
+            // but it does match the actual framework
+            Assert.Throws<ArgumentException>(() => VirtualPathUtility.ToAbsolute("hello", "/"));
+			Assert.Throws<ArgumentException>(() => VirtualPathUtility.ToAbsolute("../../test", "/"));
```
twsouthwick, line 25:
```suggestion
	Assert.Throws<ArgumentException>(() => VirtualPathUtility.ToAbsolute("../../test", "/"));
```

src/Microsoft.AspNetCore.SystemWebAdapters/VirtualPathUtility.cs
```diff
@@ -0,0 +1,698 @@
...
+
+        // Don't allow relative paths, since they cannot be made App Absolute
+        if (!IsRooted(virtualPath))
+            throw new ArgumentOutOfRangeException(nameof(virtualPath));
+
+        // Return it unchanged
+        return virtualPath;
+    }
+
+
```
twsouthwick, line 639:
```suggestion

```

src/Microsoft.AspNetCore.SystemWebAdapters/VirtualPathUtility.cs
```diff
@@ -0,0 +1,698 @@
...
+        return MakeVirtualPathAppAbsolute(virtualPath, HttpRuntime.AppDomainAppVirtualPath);
+    }
+
+    // If a virtual path is app relative (i.e. starts with ~/), change it to
+    // start with the actuall app path.
+    // E.g. ~/Sub/foo.aspx --> /MyApp/Sub/foo.aspx
+    private static string MakeVirtualPathAppAbsolute(string virtualPath, string applicationPath)
+    {
+
+        // If the path is exactly "~", just return the app root path
```
twsouthwick, line 612:
```suggestion
        // If the path is exactly "~", just return the app root path
```

src/Microsoft.AspNetCore.SystemWebAdapters/VirtualPathUtility.cs
```diff
@@ -0,0 +1,698 @@
...
+            return appRelativeCharacterString;
+
+        // Special case for apps rooted at the root:
+        if (appPathLength == 1)
+            return appRelativeCharacter + virtualPath;
+
+        return appRelativeCharacter + virtualPath[(appPathLength - 1)..];
+    }
+
+
```
twsouthwick, line 600:
```suggestion

```

twsouthwick (commented):
nit: some whitepace issues

src/Microsoft.AspNetCore.SystemWebAdapters/Ref.Standard.cs
CZEMacLeod: Hi - found a problem with the result of the script.
It brought the attribute [System.Diagnostics.CodeAnalysis.NotNullIfNotNullAttribute("virtualPath")] through for `AppendTrailingSlash` - however this doesn't exist on netstandard2.0.
I can remove the attribute from the method if required? - and it is the only method to which this actually applies.

CZEMacLeod:
> nit: some whitepace issues

I think all of these are covered in the latest update.

src/Microsoft.AspNetCore.SystemWebAdapters/Ref.Standard.cs
twsouthwick: If you pull in this commit it should work: https://github.com/twsouthwick/systemweb-adapters/commit/f166b7c4d45db18acfde280ebe25dd565694512f

CZEMacLeod: I have put in a manual work-around for now - but it feels like code smell 🫤

twsouthwick: That commit I referenced is the non-hacky way :)

test/Microsoft.AspNetCore.SystemWebAdapters.Tests/VirtualPathUtilityTests.cs
twsouthwick: This is fine

src/Microsoft.AspNetCore.SystemWebAdapters/Ref.Standard.cs
CZEMacLeod: Much better! I'm just struggling with git to try and work out how to pull in your commit to my fork!?

twsouthwick: If you add my fork as a remote and fetch the changes, you should be able to then do `git cherry-pick f166b7c`

CZEMacLeod: Cool. I think that should be that added now and reverted my hacky nastiness.

twsouthwick:
@CZEMacLeod there may be a delay in merging this per the comment here https://github.com/dotnet/systemweb-adapters/issues/83#issuecomment-1182182757. @adityamandaleeka can provide an update here

twsouthwick:
While we're waiting - do you think that #185 would help with some of the testing of non-default virtual directories?

CZEMacLeod:
@twsouthwick I'm not sure it does anything here. All the tests pass is a fixed app path rather than using the overload that defaults to the HttpRuntime value.
The difficulty is actually running these tests on full framework, rather than running them here.
Certainly, for MapPath #184, it would be useful to be able to mock AppDomainAppPath with a known value, as this would, by default, vary depending on the project path.
I didn't add any direct tests on that, as it wasn't really possible to mock httpruntime and the request, but I guess with #185 it would be possible to do this without actually running it on a project (as I do in samples/ClassLibrary/RequestInfo.cs)
I might need some help putting that together though as I'm not entirely familiar with the mocking classes/process you have.

twsouthwick:
> The difficulty is actually running these tests on full framework, rather than running them here.

I'd say that's a relatively low priority. The policy I've followed is to verify certain behavior happens on framework, and then create tests to mimic that as closely as possible using what we need to mock it up locally

> I might need some help putting that together though as I'm not entirely familiar with the mocking classes/process you have.

We've been using Moq for mocking and Autofixture for generating test data (if needed).

> I didn't add any direct tests on that, as it wasn't really possible to mock httpruntime and the request

Yup, that's the point of #185 to add the ability to do that. The VirtualPathUtility can follow a similar pattern and be an instance class internally that takes in an IHttpRuntime and is set by the IApplicationBuilder similar to the runtime.

twsouthwick:
@CZEMacLeod Can you please include the following file in your PR? https://github.com/dotnet/systemweb-adapters/blob/tasou/license/THIRD-PARTY-NOTICES.txt

Once that's in we can move forward with the PR. Thanks for your patience.

CZEMacLeod:
I don't understand why these tests failed this time round - they worked previously, and the only change was adding the text file.

twsouthwick:
Try merging locally and see if it changes things. I believe the builds merge with main and build/test that

twsouthwick (commented):
@CZEMacLeod In comparing with the framework sources, looks like you pulled in the implementation needed. Would it be much work to keep the implementation using `UrlPath`, `VirtualPath`, `StringUtil` and the other helper classes as is? Additional infrastructure we may want to bring in will be using those as well and it would be nice to keep the similar shape so as we bring in more stuff we don't start duplicating stuff.

CZEMacLeod:
> Try merging locally and see if it changes things. I believe the builds merge with main and build/test that

Ah - because #185 has been merged, the tests now fail because HttpRuntime is not set. I'll have to try and rebase after #185 and update the tests to initialize HttpRuntime.

CZEMacLeod:
> @CZEMacLeod In comparing with the framework sources, looks like you pulled in the implementation needed. Would it be much work to keep the implementation using `UrlPath`, `VirtualPath`, `StringUtil` and the other helper classes as is? Additional infrastructure we may want to bring in will be using those as well and it would be nice to keep the similar shape so as we bring in more stuff we don't start duplicating stuff.

I only brought in a small part of `UrlPath` and `StringUtil`, and didn't bring in `VirtualPath` at all - I didn't see any need for it and it seems to just result in a lot of allocations and mem copying which aren't needed. 
There have also been a number of updates/tweaks to the code afterwards.
Those classes also use resource items for error messages etc. which have all been removed.
Bringing in the whole of those classes would probably be an exercise on its own. If you do decide to bring them in, in the future, you could check if any of the functions in `VirtualPathUtility` match and swap out to use the shared code.
For now, I would prefer to keep this PR limited to the original scope of implementing just that class.

twsouthwick:
> I only brought in a small part of UrlPath and StringUtil, and didn't bring in VirtualPath at all - I didn't see any need for it and it seems to just result in a lot of allocations and mem copying which aren't needed.

Yeah, the challenge is that a lot of stuff internally ends up relying on these methods/classes. It will simplify future work to bring stuff in in the same shape and then we can deal with any perf impacts en masse (i.e. VirtualPath could be converted to use spans/be a struct/etc). I'd really prefer that this happens as part of this PR, otherwise, I'll probably end up needing to redo it to support that scenario. No need to bring in the whole classes - just the parts you used.

When you say "didn't bring in VirtualPath" at all, you mean you inlined it, right? All the implementation at https://referencesource.microsoft.com/#System.Web/VirtualPathUtility.cs,2b829f9059f552ca shows things being delegated to `VirtualPath`/`StringUtil`/`UrlPath` for the work. Those methods that are required are what I'd like to see brought in with this PR (nothing beyond what is needed - that easily can balloon beyond what the original scope here is).

> Merge weirdness

Looks like your merge may have done something weird. It's showing files that are in the project already as being additions. Maybe try rebasing and force pushing?

CZEMacLeod:
@twsouthwick The implementation of `VirtualPathUtility` is partly my own, based on the utility methods. It doesn't use `VirtualPath` at all, nor did I inline any of that. I did move some of the code up from the `UrlPath` and `StringUtil` classes up to the main methods when nothing else was needed, but skipped all the `VirtualPath` stuff entirely. It is mainly using some of the utility methods combined with implementing what the docs say.
I could certainly move some of the support methods back into those classes, although - e.g. for the `MapPath` implementation in #184 it just uses the functions of `VirtualPathUtility` directly rather than hitting the underlying stuff.

As for the merge stuff - yeah. I had to pull in the changes from #185 due to the changes to (I)HttpRuntime, then it thought there were merge conflicts for a file I didn't touch, so I brought in the other changes from #186 so it would merge correctly. I think it is fine now, but I can try rebasing it again if needed.

twsouthwick:
Ok - I think I understand now. For now, then, can you put the methods you pulled from `UrlPath` and `StringUtil` back into static class with those names  we can skip the `VirtualPath` stuff for now?

> Merge stuff

I'm still seeing a bunch of erroneous changes. It may be GitHub showing the diff weird, but can you try rebasing?

CZEMacLeod:
I'm looking at breaking out the `UrlPath` stuff and found that some of the methods rely directly or indirectly on `HttpRuntime` - we had previously discussed potentially making `VirtualPathUtility` an instance class like `HttpRuntime` and injecting `IHttpRuntime`. This is definitely going to make that process more complicated.
There is only a single method in `StringUtil` that is used - `StringStartsWithIgnoreCase`.

CZEMacLeod:
@twsouthwick I've broken out the methods into the util classes, and rebased. Hopefully this will a) merge correctly without whatever weirdness you were seeing and b) build, and test correctly (it does locally).

src/Microsoft.AspNetCore.SystemWebAdapters/Utilities/StringUtil.cs
```diff
@@ -0,0 +1,15 @@
+namespace System.Web.Util;
```
twsouthwick, line 1: nit: these files need the header. Apparently we don't have a code fix for that :(

twsouthwick:
> Hopefully this will a) merge correctly without whatever weirdness you were seeing and b) build, and test correctly (it does locally).

Yup on both accounts :) I've seen that weirdness with GH when doing merges. Not exactly sure why, but rebasing seems to fix it.

src/Microsoft.AspNetCore.SystemWebAdapters/Utilities/StringUtil.cs
CZEMacLeod: Ah - Yeah - should probably have put the original source mention like I did in VirtualPathUtility.cs.
I did tweak `StringUtl.StringStartsWithIgnoreCase` to use the newer StartsWith overload (I assume it was written against a much earlier version of .Net and never changed). I did do a quick BenchmarkDotNet test of the original code vs my version and got a x2 performance improvement. It still needs the wrapper due to different exception handing for null values.

CZEMacLeod: Added attribution for the classes, and the header as per the other files.

Tratcher (approved):

src/Microsoft.AspNetCore.SystemWebAdapters/Utilities/UrlPath.cs
```diff
@@ -0,0 +1,496 @@
...
+    internal const char AppRelativeCharacter = '~';
+    internal const string AppRelativeCharacterString = "~/";
+    private const string Invalid_vpath = "'{0}' is not a valid virtual path.";
+    private const string Physical_path_not_allowed = "'{0}' is a physical path, but a virtual path was expected.";
+    private const string Cannot_exit_up_top_directory = "Cannot use a leading .. to exit above the top directory.";
+    internal const string Path_must_be_rooted = "The virtual path '{0}' is not rooted.";
+
+    internal static bool HasTrailingSlash(string virtualPath) => virtualPath[^1] == '/';
+
+    internal static bool IsRooted(string basepath) => string.IsNullOrEmpty(basepath) || basepath[0] == '/' || basepath[0] == '\\';
```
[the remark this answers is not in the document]
CZEMacLeod: It is according to the original code - https://github.com/microsoft/referencesource/blob/master/System.Web/Util/UrlPath.cs#L33
Null or Empty tends to get caught further up in the calls in VirtualPathUtility and either returns something explicit or throws.

src/Microsoft.AspNetCore.SystemWebAdapters/Utilities/UrlPath.cs
```diff
@@ -0,0 +1,496 @@
...
+    }
+
+    internal static string Combine(string appPath, string basepath, string relative)
+    {
+        string path;
+
+        if (string.IsNullOrEmpty(relative))
+            throw new ArgumentNullException(nameof(relative));
+        if (string.IsNullOrEmpty(basepath))
+            throw new ArgumentNullException(nameof(basepath));
```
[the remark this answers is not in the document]
twsouthwick: @Tratcher should we add stylecop to the repo so this is enforced?

src/Microsoft.AspNetCore.SystemWebAdapters/Utilities/UrlPath.cs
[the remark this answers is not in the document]
twsouthwick: FYI These methods are taken as-is (mostly) from System.Web and is how they handled things. Not sure how well we can reason about some of these things :)

src/Microsoft.AspNetCore.SystemWebAdapters/Utilities/UrlPath.cs
```diff
@@ -0,0 +1,496 @@
...
+            return true;
+        return false;
+    }
+
+    private static bool IsAbsolutePhysicalPath(string path)
+    {
+        if (path == null || path.Length < 3) return false;
+
+        // e.g c:\foo
+        if (path[1] == ':' && IsDirectorySeparatorChar(path[2])) return true;
```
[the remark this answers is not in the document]
CZEMacLeod: Again - this is based on the original code. I'm not sure there is ever a reason that something like `\:\` would be allowed as a virtual path - the use of this function is to throw if it is __not__ a valid virtual path in `CheckValidVirtualPath`.

src/Microsoft.AspNetCore.SystemWebAdapters/Utilities/UrlPath.cs
```diff
@@ -0,0 +1,507 @@
...
+
+        // Single character case: "~"
+        if (len == 1) return true;
+
+        // If it's longer, checks if it starts with "~/" or "~\"
+        return virtualPath[1] == '\\' || virtualPath[1] == '/';
+    }
+
+    // Same as Reduce, but for a virtual path that is known to be well formed
+    internal static string ReduceVirtualPath(string path)
```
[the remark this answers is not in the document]
CZEMacLeod: Unless the original call did, I don't think so. This is mainly used with the result of `MakeVirtualPathAppAbsolute` or `MakeRelative` which would normally be called with hardcoded strings from the developer, and possibly the current request path, which I think should also be 'clean'.
As it says in the description - only use on a virtual path that is known to be well formed.
This code is pretty much untouched from the original .Net implementation, other than some changes to substring handling and error messages.
If it needs to be made more robust than the original, then I think that is a separate issue?

src/Microsoft.AspNetCore.SystemWebAdapters/Utilities/UrlPath.cs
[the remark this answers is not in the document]
CZEMacLeod: Yes - `UrlPath` and `StringUtil` (other than code optimizations based on newer c# / .net functions and error strings) are pretty much a direct port.

src/Microsoft.AspNetCore.SystemWebAdapters/Utilities/UrlPath.cs
```diff
@@ -0,0 +1,496 @@
...
+            // If the base path includes a file name, get rid of it before combining
+            var lastSlashIndex = basepath.LastIndexOf('/');
+            Debug.Assert(lastSlashIndex >= 0);
+            if (lastSlashIndex < basepath.Length - 1)
+            {
+                basepath = basepath[..(lastSlashIndex + 1)];
+            }
+        }
+
+        // Make sure it's a virtual path (ASURT 73641)
```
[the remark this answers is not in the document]
CZEMacLeod: Done

src/Microsoft.AspNetCore.SystemWebAdapters/Utilities/UrlPath.cs
```diff
@@ -0,0 +1,496 @@
...
+
+        var virtualPath2Length = virtualPath2.Length;
+
+        // same length - same path
+        if (virtualPath1.Length == virtualPath2Length)
+        {
+            return true;
+        }
+
+        // Special case for apps rooted at the root. VSWhidbey 286145
```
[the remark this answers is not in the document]
CZEMacLeod: Done

PR #183, opened 2022-08-29
Remove APIs that are not implemented

twsouthwick:
This was a placeholder, but not planned for v1, so we should just remove it. If something is not implemented, it shouldn't be available.

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters/Constants.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters/Constants.cs
@@ -5,7 +5,5 @@ namespace System.Web;
 
 internal static class Constants
 {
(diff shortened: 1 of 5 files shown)


CZEMacLeod:
I just did a MapPath implementation based on VirtualPathProvider - I'm not sure how much testing I can do for this - I added some output to RequestInfo to show the data, and it was the same between MvcApp and MvcCoreApp (except the fact that they are in different folders).
If you want it, you are welcome to it 'as-is' https://github.com/CZEMacLeod/systemweb-adapters/tree/czemacleod/mappath

twsouthwick:
@CZEMacLeod awesome! I'll postpone merging this if you want to create a PR for that

twsouthwick:
@CZEMacLeod is implementing these in #184

PR #184, opened 2022-08-29
Implement MapPath and AppDomainAppPath

CZEMacLeod:
Implement HttpRuntime.AppDomainAppPath, HttpServerUtility.MapPath, and HttpServerUtilityWrapper.MapPath

**Provide support for MapPath**
Paths are mapped relative the application root directory - available as HttpRuntime.AppDomainAppPath

This may not behave as expected, for example if a person uses MapPath to write uploaded files to disk and expects them to be served, due to the way AspNetCore uses ContentRootPath/WebRootPath.
In this case it might be better to implement your own helper function which can use something like
```cs
namespace System.Web.Extensions
{
    public static class MapPathExtension
    {
        public static string MapContentPath(this System.Web.HttpContext context, string? path)
        {
#if NET6_0_OR_GREATER
            var appPath = path is null ? VirtualPathUtility.GetDirectory(context.Request.Path) :
                VirtualPathUtility.Combine(
                VirtualPathUtility.GetDirectory(context.Request.Path) ?? "/"
                , path);
            var coreContext = (Microsoft.AspNetCore.Http.HttpContext)context;
            var rootPath = coreContext.RequestServices.GetRequiredService<IWebHostEnvironment>().WebRootPath;
            if (string.IsNullOrEmpty(appPath)) return rootPath;
            return System.IO.Path.Combine(rootPath, appPath[1..].Replace('/', System.IO.Path.DirectorySeparatorChar));
#else
            return context.Server.MapPath(path);
#endif
        }
    }
}
```

Addresses #183

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpRuntimeFactory.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpRuntimeFactory.cs
@@ -24,7 +24,7 @@ internal static class HttpRuntimeFactory
     {
         public string AppDomainAppVirtualPath => "/";
 
-        public string AppDomainAppPath => AppDomain.CurrentDomain.BaseDirectory;
+        public string AppDomainAppPath => AppContext.BaseDirectory;
     }
 
     internal class IISHttpRuntime : IHttpRuntime
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpRuntimeFactory.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpRuntimeFactory.cs
@@ -1,6 +1,7 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System;
 using Microsoft.AspNetCore.SystemWebAdapters;
 
 namespace Microsoft.Extensions.DependencyInjection;
@@ -22,6 +23,8 @@ internal static class HttpRuntimeFactory
     internal class DefaultHttpRuntime : IHttpRuntime
     {
         public string AppDomainAppVirtualPath => "/";
+
+        public string AppDomainAppPath => AppDomain.CurrentDomain.BaseDirectory;
     }
 
     internal class IISHttpRuntime : IHttpRuntime
@@ -34,5 +37,7 @@ internal static class HttpRuntimeFactory
         }
 
         public string AppDomainAppVirtualPath => _config.pwzVirtualApplicationPath;
+
+        public string AppDomainAppPath => _config.pwzFullApplicationPath;
     }
 }
--- a/src/Microsoft.AspNetCore.SystemWebAdapters/HttpServerUtility.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters/HttpServerUtility.cs
@@ -1,7 +1,9 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.WebUtilities;
+using Microsoft.Extensions.DependencyInjection;
 
 namespace System.Web;
 
@@ -21,10 +23,14 @@ public class HttpServerUtility
     {
         var appPath = path is null ? VirtualPathUtility.GetDirectory(_context.Request.Path) :
             VirtualPathUtility.Combine(
-            VirtualPathUtility.GetDirectory(_context.Request.Path)
+            VirtualPathUtility.GetDirectory(_context.Request.Path) ?? "/"
             , path);
-        if (string.IsNullOrEmpty(appPath)) return HttpRuntime.AppDomainAppPath;
-        return System.IO.Path.Combine(HttpRuntime.AppDomainAppPath, appPath[1..].Replace('/', System.IO.Path.DirectorySeparatorChar));
+        // Potential alternative implementations here - for the most literal match to the original use HttpRuntime.AppDomainAppPath
+        var rootPath = HttpRuntime.AppDomainAppPath;
+        //var rootPath = _context.RequestServices.GetRequiredService<IWebHostEnvironment>().WebRootPath;
+        //var rootPath = _context.RequestServices.GetRequiredService<IWebHostEnvironment>().ContentRootPath;
+        if (string.IsNullOrEmpty(appPath)) return rootPath;
+        return System.IO.Path.Combine(rootPath, appPath[1..].Replace('/', System.IO.Path.DirectorySeparatorChar));
     }
 
     [Diagnostics.CodeAnalysis.SuppressMessage("Design", "CA1024:Use properties where appropriate", Justification = Constants.ApiFromAspNet)]
--- a/src/Microsoft.AspNetCore.SystemWebAdapters/HttpServerUtility.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters/HttpServerUtility.cs
@@ -17,8 +17,15 @@ public class HttpServerUtility
 
     public string MachineName => Environment.MachineName;
 
-    [Obsolete(Constants.NotImplemented)]
-    public string MapPath(string path) => throw new NotImplementedException();
+    public string MapPath(string? path)
+    {
+        var appPath = path is null ? VirtualPathUtility.GetDirectory(_context.Request.Path) :
+            VirtualPathUtility.Combine(
+            VirtualPathUtility.GetDirectory(_context.Request.Path)
+            , path);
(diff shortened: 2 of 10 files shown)


twsouthwick:
@CZEMacLeod this builds on #179, right? If so, I'll hold off on reviewing until we get that in.

CZEMacLeod:
@twsouthwick Yes - It uses the `VirtualPathUtility` from #179 to handle the app relative and request relative paths and prevent 'exiting' up above the site directory.

CZEMacLeod:
@twsouthwick The tests I've added seem to be failing on Linux/Mac - I think because the expected results rely on the directory separator character but I'm not sure how to find the output, or what to do to mitigate this...

src/Microsoft.AspNetCore.SystemWebAdapters/HttpServerUtility.cs
```diff
@@ -17,8 +19,19 @@ internal HttpServerUtility(HttpContextCore context)
...
 
-    [Obsolete(Constants.NotImplemented)]
-    public string MapPath(string path) => throw new NotImplementedException();
+    public string MapPath(string? path)
+    {
+        var appPath = (string.IsNullOrEmpty(path) ? VirtualPathUtility.GetDirectory(_context.Request.Path) :
+            VirtualPathUtility.Combine(
+                VirtualPathUtility.GetDirectory(_context.Request.Path) ?? "/"
+                , path));
+        // Potential alternative implementations here - for the most literal match to the original use HttpRuntime.AppDomainAppPath
```
twsouthwick, line 28: How should we decide the approach here? What are the pros and cons? I'm wondering if this is something we'd want to provide options to toggle between (either globally or per endpoint).

test/Microsoft.AspNetCore.SystemWebAdapters.Tests/HttpServerUtilityTests.cs
```diff
@@ -77,4 +86,53 @@ public void UrlTokenRoundtrip(string input, string expected)
...
+    [InlineData("/RootLevelPage.aspx", "/NotRealFolder", "C:\\ExampleSites\\TestMapPath\\NotRealFolder")]
+    [InlineData("/DownOneLevel/DownLevelPage.aspx", null, "C:\\ExampleSites\\TestMapPath\\DownOneLevel")]
+    [InlineData("/DownOneLevel/DownLevelPage.aspx", "../RootLevelPage.aspx", "C:\\ExampleSites\\TestMapPath\\RootLevelPage.aspx")]
+    [InlineData("/api/test/request/info", null, "C:\\ExampleSites\\TestMapPath\\api\\test\\request")]
+    [InlineData("/api/test/request/info", "", "C:\\ExampleSites\\TestMapPath\\api\\test\\request")]
+    [InlineData("/api/test/request/info", "/UploadedFiles", "C:\\ExampleSites\\TestMapPath\\UploadedFiles")]
+    [InlineData("/api/test/request/info", "UploadedFiles", "C:\\ExampleSites\\TestMapPath\\api\\test\\request\\UploadedFiles")]
+    [InlineData("/api/test/request/info", "~/MyUploadedFiles", "C:\\ExampleSites\\TestMapPath\\MyUploadedFiles")]
+    [Theory]
+    public void MapPath(string page, string? path, string expected)
```
twsouthwick, line 103: since the outcome here is expected to be a valid path, it makes sense that the tests are failing on linux/mac. 

What about changing the test to the following:

```suggestion
    public void MapPath(string page, string? path, params string[] expectedSegment)
```

Then using `Path.Combine` to construct the expected string?

src/Microsoft.AspNetCore.SystemWebAdapters/HttpServerUtility.cs
CZEMacLeod: I was thinking maybe an extension method/overload for `MapPath` that allows passing the root path, (or an enum value) to say which scenario you want. This would allow you to use your existing code as-is, but use the overload on a per-use case scenario. If it was available to both Core and Framework, you could adjust your code as necessary and still have it work on both sides (e.g. from a class library).

I could also see adding some kind of IMapPathRoot to the services where you could handle it as you wish. This could also maybe match up with something from StaticFiles / PhysicalFileProvider / CompositeFileProvide if you are mapping multiple folders?

I'm really not sure the best approach here - and the current implementation follows the original System.Web approach of mapping directly to the 'website root' just now.

I guess a user could also create their own `IHttpRuntime` with the valid path for their use case (although maybe in that case we should make `IISHttpRuntime` public and AppDomainAppPath virtual? (Kind of like the tests do just now)

test/Microsoft.AspNetCore.SystemWebAdapters.Tests/HttpServerUtilityTests.cs
CZEMacLeod: I've implemented this (as well as adjusting the `AppDomainAppPath` based on `System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform(OSPlatform.Windows)`).
It would be really nice to be able to run the xunit tests using WSL and/or docker to be able to check the Linux tests on a windows machine. I think this is possible, but might require some setup to do.

src/Microsoft.AspNetCore.SystemWebAdapters/HttpServerUtility.cs
CZEMacLeod: I'm not sure how well it would work, but here is (yet) another implementation:
```cs
        var appPath = (string.IsNullOrEmpty(path) ? VirtualPathUtility.GetDirectory(_context.Request.Path) :
            VirtualPathUtility.Combine(
                VirtualPathUtility.GetDirectory(_context.Request.Path) ?? "/"
                , path));
        var rootProvider = _context.RequestServices.GetRequiredService<IWebHostEnvironment>().WebRootFileProvider;
        return rootProvider.GetFileInfo(appPath).PhysicalPath;
```
This might work for the various scenarios listed in https://docs.microsoft.com/en-us/aspnet/core/fundamentals/static-files?view=aspnetcore-6.0 like [Serve files from multiple locations](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/static-files?view=aspnetcore-6.0#serve-files-from-multiple-locations).

twsouthwick: Hmm, yeah there's definitely a number of directions we can go here.  For now, we can just go with the content root as the default (that matches the application path in HttpRuntime).

We'll need to consider how to expose a way to override it through an API review. We're locking down the APIs for v1, so we'll probably wait for any non-System.Web APIs (i.e. things like IMapPath) to be enabled after we get v1 shipped.

If the content path is sufficient for now, we can move forward with the PR. Otherwise, we may want to hold off until after v1.

CZEMacLeod: I think for initial use, the implementation that is in the code just now is fine. There are use cases beyond the basics that System.Web supports natively (which is why I have IMapPath in my projects anyway), that could be implemented, but I don't think it is _required_ for this to be usable.

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpRuntimeFactory.cs
```diff
@@ -22,6 +23,8 @@ public static IHttpRuntime Create()
     internal class DefaultHttpRuntime : IHttpRuntime
     {
         public string AppDomainAppVirtualPath => "/";
+
+        public string AppDomainAppPath => AppDomain.CurrentDomain.BaseDirectory;
```
twsouthwick, line 27:
```suggestion
        public string AppDomainAppPath => AppContext.BaseDirectory;
```

twsouthwick (changes requested):
Can you clean up the comments and such in the `MapPath` implementation to just have the direction we decided on?

test/Microsoft.AspNetCore.SystemWebAdapters.Tests/HttpServerUtilityTests.cs
CZEMacLeod: Woot - All checks passed!

twsouthwick (approved):

twsouthwick:
Thanks @CZEMacLeod for the contribution! LGTM

src/Microsoft.AspNetCore.SystemWebAdapters/HttpServerUtility.cs
CZEMacLeod: I've placed some initial work for `IMapPath` and the different implementations at https://github.com/CZEMacLeod/systemweb-adapters/tree/czemacleod/imappath
This is loosely based on the discussions above, combined with some Framework implementation I already have in production.
Feel free to see if any of it would be of interest - although as you say, this is probably a post V1 thing.

Tratcher (approved):

PR #185, opened 2022-08-30
Add a singleton IHttpRuntime to abstract runtime data

twsouthwick:
Currently, anything that relies on HttpRuntime ends up relying on static data that is initialized differently depending on if you're on IIS or not. This change introduces an abstraction so things that require HttpRuntime can build on it in a testable way.


samples/ClassLibrary/RequestInfo.cs
```diff
@@ -13,6 +13,7 @@ public static void WriteRequestInfo(bool suppress)
         using (var writer = new SimpleJsonWriter(context.Response))
         {
             writer.Write("VirtualDirectory", HttpRuntime.AppDomainAppVirtualPath);
+            writer.Write("VirtualDirectory", context.Request.ApplicationPath);
```
mjrousos, line 16: Nit-pick - same name as the above line could be a little confusing

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/SystemWebAdaptersExtensions.cs
```diff
@@ -17,6 +16,7 @@ public static class SystemWebAdaptersExtensions
     public static ISystemWebAdapterBuilder AddSystemWebAdapters(this IServiceCollection services)
     {
         services.AddHttpContextAccessor();
+        services.AddSingleton<IHttpRuntime>(_ => HttpRuntimeFactory.Create());
```
mjrousos, line 19:
```suggestion
        services.AddSingleton<IHttpRuntime>(HttpRuntimeFactory.Create);
```

mjrousos (approved):

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/SystemWebAdaptersExtensions.cs
twsouthwick: That wouldn't work - there's no overload that takes a `Func<T>` but rather it expects a `Func<IServiceProvider,T>`

mjrousos: Ah, good point. I missed that it expects a service provider.

Tratcher (approved):

HaoK (approved):

PR #186, opened 2022-08-30
Add Content-Type as default header that is set

twsouthwick:
This was a default value that it appears some internal teams had relied on and caused problems when they were migrating. This change expands an existing middleware to handle adding any default response headers


src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/SetDefaultResponseHeadersMiddleware.cs
```diff
@@ -0,0 +1,37 @@
...
+        (HeaderNames.ContentType, "text/html")
+    };
+
+    public SetDefaultResponseHeadersMiddleware(RequestDelegate next) => _next = next;
+
+    public Task InvokeAsync(HttpContext context)
+    {
+        foreach (var (header, value) in _defaultHeaders)
+        {
+            if (!context.Response.Headers.ContainsKey(header))
```
mjrousos, line 29: Since we only set headers that aren't already present, should we call `_next` first so that whatever endpoint handles the request has an opportunity to write headers and we'll only set the default ones if they're still unset after that?

mjrousos (approved):
LGTM. One question about whether it makes sense to set these headers before or after subsequent middleware run, but either way likely works.

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/SetDefaultResponseHeadersMiddleware.cs
twsouthwick: We can't really do that because after `_next(context)`  the response will have started writing. The alternative would be to register a  `Response.OnStarting` delegate that would handle it there. Let me think about the implications of either.

mjrousos: That's a good point. I'm ok with keeping this model if it's simpler.

PR #192, opened 2022-09-02
Handle application paths for remote calls

### LINKED ISSUE #191: Remote authentication not working when Framework app is not the root of the website
I am trying to get the remote authentication working against our ASP.NET MVC 5.2 app. Our app doesn't run in the root of the IIS website - we have to run it as a sub application (i.e. /myapp), and I think this may be causing the problem I am having. Essentially, when the Core app makes the authentication request to the Framework app, it calls "/systemweb-adapters/authenticate" and not "/myapp/systemweb-adapters/authenticate", which then fails with a 404 (I can see this in the IIS logs).

In Program.cs I have the following:

```
var builder = WebApplication.CreateBuilder(args);
    .AddRemoteApp(options =>
    {
        options.RemoteAppUrl = new("http://localhost/MyApp");
        options.ApiKey = "MySecretKey";
    })
    .AddRemoteAppAuthentication(true);
```

But it seems to ignore the "/MyApp" bit when it makes the request. I have created a dummy solution with a framework app and a core app, and can get the remote authentication working just fine when the framework app is in the root of a website, but as soon as I put it in a sub folder, it fails due to the above path problem.

Is this a bug, or I am approaching this incorrectly?

twsouthwick:
This change does two things:

1) Ensures that the root address ends in a '/' and that the request path starts with a './'. This is done in a way that gets cached and won't be recomputed each request
2) Update the RemoteModule to compare the incoming strings against the app relative path that accounts for any application paths or virtual directories.

Fixes #191

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/RemoteAppClientExtensions.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/RemoteAppClientExtensions.cs
@@ -3,6 +3,8 @@
 
 using System;
 using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.AspNetCore.SystemWebAdapters;
 using Microsoft.Extensions.Options;
 
@@ -29,6 +31,7 @@ public static class RemoteAppClientExtensions
         builder.Services.AddOptions<RemoteAppClientOptions>()
             .ValidateDataAnnotations();
 
+        builder.Services.AddTransient<HandleVirtualDirectoryHandler>();
         builder.Services.AddHttpClient(RemoteConstants.HttpClientName)
             .ConfigurePrimaryHttpMessageHandler(sp =>
             {
@@ -42,6 +45,7 @@ public static class RemoteAppClientExtensions
                 // Disable cookies in the HTTP client because the service will manage the cookie header directly
                 return new HttpClientHandler { UseCookies = false, AllowAutoRedirect = false };
             })
+            .AddHttpMessageHandler<HandleVirtualDirectoryHandler>()
             .ConfigureHttpClient((sp, client) =>
             {
                 var options = sp.GetRequiredService<IOptions<RemoteAppClientOptions>>().Value;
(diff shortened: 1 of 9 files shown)


src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/RemoteAppClientExtensions.cs
```diff
@@ -73,6 +77,37 @@ public static ISystemWebAdapterRemoteClientAppBuilder Configure(this ISystemWebA
         return builder;
     }
 
+    /// <summary>
+    /// <see cref="HttpClient.BaseAddress"/> is set automatically, but if this is supposed to have a path, then that gets
+    /// lost. This handler will append that back if necessary.
```
Tratcher, line 82: There's a way to make the request that avoids this issue. It will take some explaining, let's talk offline to see if it's any easier than this.

Tratcher (approved):

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/RemoteAppClientExtensions.cs
```diff
@@ -3,6 +3,8 @@
 
 using System;
 using System.Net.Http;
+using System.Threading;
```
HaoK, line 6: nit unused since only change was adding using?

test/Microsoft.AspNetCore.SystemWebAdapters.CoreServices.Tests/RemoteAppClientOptionsTests.cs
```diff
@@ -35,7 +35,11 @@ public void VerifyIsCalled(string apiKeyHeader, string apiKey, string remoteAppU
             {
                 options.ApiKey = apiKey;
                 options.ApiKeyHeader = apiKeyHeader;
-                options.RemoteAppUrl = (remoteAppUrl is null ? null : new Uri(remoteAppUrl, UriKind.Absolute))!;
+
+                if (remoteAppUrl is not null)
```
HaoK, line 39: Is this all that's needed to ensure we can handle the new application paths in this PR?

twsouthwick: no - that just fixes a test failure. The the application paths fix is in the "RemoteModule" changes

PR #194, opened 2022-09-07
Add a playwright test project for MvcApp

HaoK:
Initial basic test that verifies login/logout works across Mvc[Core]App using playwright

Currently both apps need to be started externally before running dotnet test on the playwright tests, the current idea is that we will ask CTI to run these tests once a month

### DIFF
--- a/test/Samples.MVCApp.Tests/MVCSampleTest.cs
+++ b/test/Samples.MVCApp.Tests/MVCSampleTest.cs
@@ -1,16 +1,20 @@
 using Microsoft.Playwright.NUnit;
+using NUnit.Framework;
 
 namespace PlaywrightTests;
 
 [Parallelizable(ParallelScope.Self)]
 public class MvcSampleTest : PageTest
 {
+    private const string MvcAppUrl = "https://localhost:44339/";
+    private const string MvcCoreAppUrl = "https://localhost:55442/";
+
     [Test]
     public async Task MVCCoreAppCanLogoutBothApps()
     {
         var email = $"{Path.GetRandomFileName()}@test.com";
 
-        await Page.GotoAsync("https://localhost:44339/");
+        await Page.GotoAsync(MvcAppUrl);
         await Expect(Page.Locator("text=My ASP.NET Application")).ToBeVisibleAsync();
 
         // Create the user
@@ -22,13 +26,13 @@ public class MvcSampleTest : PageTest
         await Expect(Page.Locator($"text=Hello {email}!")).ToBeVisibleAsync();
 
         // Make sure core app also logged in
-        await Page.GotoAsync("https://localhost:55442/");
+        await Page.GotoAsync(MvcCoreAppUrl);
         await Expect(Page.Locator($"text=Hello {email}!")).ToBeVisibleAsync();
 
         // Logout on core app and make sure both logged out
         await Page.Locator(@"text=Log out").ClickAsync();
         await Expect(Page.Locator(@"text=Log in")).ToBeVisibleAsync();
-        await Page.GotoAsync("https://localhost:44339/");
+        await Page.GotoAsync(MvcAppUrl);
         await Expect(Page.Locator(@"text=Log in")).ToBeVisibleAsync();
     }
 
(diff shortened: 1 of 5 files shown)


test/Samples.MVCApp.Tests/Samples.MVCApp.Tests.csproj
```diff
@@ -0,0 +1,20 @@
...
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.2.0" />
+    <PackageReference Include="Microsoft.Playwright.NUnit" Version="1.25.0" />
+    <PackageReference Include="NUnit" Version="3.13.3" />
```
twsouthwick, line 14: We're using xunit elsewhere in the project. Is there a need to use NUnit?

HaoK: I just followed the playwright docs, I can see if they have xunit versions

HaoK: Doesn't look like there's as mature built in support for xunit currently, as there are lots of projects, i.e. https://github.com/martincostello/dotnet-playwright-tests which seem to be trying to address that integration, I think we should consider just using the playwright nunit (or mstest packages) rather than taking on the extra complexity of trying to make it work with xunit?  https://playwright.dev/dotnet/docs/test-runners#xunit-support

HaoK: > While Playwright for .NET isn't tied to a particular test runner or testing framework, in our experience it works best with the built-in .NET test runner, and using NUnit as the test framework. NUnit is also what we use internally for [our tests](https://github.com/microsoft/playwright-dotnet/tree/main/src/Playwright.Tests).

twsouthwick: That's reasonable. I'm not familiar with nunit, but I'm assuming it's got similar constructs :)

test/Samples.MVCApp.Tests/MVCSampleTest.cs
```diff
@@ -0,0 +1,62 @@
...
+
+[Parallelizable(ParallelScope.Self)]
+public class MvcSampleTest : PageTest
+{
+    [Test]
+    public async Task MVCCoreAppCanLogoutBothApps()
+    {
+        var email = $"{Path.GetRandomFileName()}@test.com";
+
+        await Page.GotoAsync("https://localhost:44339/");
```
twsouthwick, line 13: should these urls be some sort of const/options, i.e. FrameworkApp, CoreApp

test/Samples.MVCApp.Tests/Samples.MVCApp.Tests.csproj
```diff
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net7.0</TargetFramework>
```
twsouthwick, line 4: nit: other tests are currently .net 6

twsouthwick (approved):

test/Samples.MVCApp.Tests/Usings.cs
```diff
@@ -0,0 +1 @@
+global using NUnit.Framework;
```
twsouthwick, line 1: is this needed? isn't this an auto-generated file?

twsouthwick:
An additional thought - there are other test apps that exercise different auth types. Are you planning on adding those to the CTI test line up as well?

HaoK:
Sure I can do that, can you list which flows are the most relevant for CTI to cover and I can do a separate PR adding those new tests too

twsouthwick:
They're the sample apps in https://github.com/dotnet/systemweb-adapters/tree/main/samples/RemoteAuth

HaoK:
I'll add playwright tests for the other sample remote auth apps in a follow up PR

CZEMacLeod:
Since this pr was added I'm getting the following error:
`Samples.MVCApp.Tests\Samples.MVCApp.Tests.csproj : error NU1101: Unable to find package Microsoft.Playwright.NUnit. No packages exist with this id in source(s): C:\Program Files\dotnet\library-packs, dotnet-eng, dotnet-public, dotnet-tools, dotnet6, myget-legacy`
Is this 'expected'?
I got round it by adding `<add key="nuget.org" value="https://api.nuget.org/v3/index.json" />` to `nuget.config` but not sure how this works on the build servers?

PR #196, opened 2022-09-08
Add remote auth forms test

HaoK:
Playwright tests for RemoteAuth/forms

### DIFF
--- a/test/Samples.RemoteAuth.Forms.Tests/FormsSampleTest.cs
+++ b/test/Samples.RemoteAuth.Forms.Tests/FormsSampleTest.cs
@@ -1,0 +1,66 @@
+using Microsoft.Playwright.NUnit;
+using NUnit.Framework;
+
+namespace PlaywrightTests;
+
+[Parallelizable(ParallelScope.Self)]
+public class FormsSampleTest : PageTest
+{
+    private const string FrameworkAppUrl = "https://localhost:44394/";
+    private const string CoreAppUrl = "https://localhost:7080/";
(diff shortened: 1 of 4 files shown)


test/Samples.RemoteAuth.Forms.Tests/FormsSampleTest.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        // Make sure core app also logged in
+        await Page.GotoAsync(CoreAppUrl);
+        await Expect(Page.Locator($"text=Hello {username}!")).ToBeVisibleAsync();
+
+        // Logout on core app and make sure both logged out
+        await Page.Locator(@"text=Log out").ClickAsync();
+        await Expect(Page.Locator($"text=You have been logged out")).ToBeVisibleAsync();
+
+        // Note: Logout on core app doesn't logout framework app
```
mjrousos, line 35: I thought logging out on the core app did log out on Framework, or did we not update the WebForms sample to do that?

mjrousos (approved):
LGTM

twsouthwick (approved):

test/Samples.RemoteAuth.Forms.Tests/FormsSampleTest.cs
HaoK: I didn't update the webforms sample and it would be different since webforms isn't using dataprotection/owin which is the crux of how the mvc apps can do things easily

PR #198, opened 2022-09-12
API updates per review feedback

twsouthwick:
https://github.com/dotnet/aspnetcore/issues/42091

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/RemoteAppClientExtensions.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/RemoteAppClientExtensions.cs
@@ -14,7 +14,7 @@ public static class RemoteAppClientExtensions
     /// Add configuration for connecting to a remote app for System.Web extensions that require a remote
     /// ASP.NET app such as remote app authentication or remote app session sharing.
     /// </summary>
-    public static ISystemWebAdapterBuilder AddRemoteAppClient(this ISystemWebAdapterBuilder builder, Action<ISystemWebAdapterRemoteClientAppBuilder> configure)
+    public static ISystemWebAdapterRemoteClientAppBuilder AddRemoteAppClient(this ISystemWebAdapterBuilder builder, Action<RemoteAppClientOptions> configure)
     {
         if (builder is null)
         {
@@ -27,6 +27,7 @@ public static class RemoteAppClientExtensions
         }
 
         builder.Services.AddOptions<RemoteAppClientOptions>()
+            .Configure(configure)
             .ValidateDataAnnotations();
 
         builder.Services.AddHttpClient(RemoteConstants.HttpClientName)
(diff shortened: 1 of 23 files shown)


src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/RemoteAppClientExtensions.cs
```diff
@@ -50,27 +51,7 @@ public static ISystemWebAdapterBuilder AddRemoteAppClient(this ISystemWebAdapter
...
-    }
-
-    public static ISystemWebAdapterRemoteClientAppBuilder Configure(this ISystemWebAdapterRemoteClientAppBuilder builder, Action<RemoteAppClientOptions> configure)
-    {
-        if (builder is null)
-        {
-            throw new ArgumentNullException(nameof(builder));
-        }
-
-        if (configure is null)
```
mjrousos, line 65: Should probably keep this error check

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/RemoteAppServerExtensions.cs
```diff
@@ -25,29 +25,10 @@ public static ISystemWebAdapterBuilder AddRemoteAppServer(this ISystemWebAdapter
...
-    }
-
-    public static ISystemWebAdapterRemoteServerAppBuilder Configure(this ISystemWebAdapterRemoteServerAppBuilder builder, Action<RemoteAppServerOptions> configure)
-    {
-        if (builder is null)
-        {
-            throw new ArgumentNullException(nameof(builder));
-        }
-
-        if (configure is null)
```
mjrousos, line 42: Again, probably want to keep the null check

mjrousos (approved):
LGTM. A few minor comments inline.

Tratcher (approved):

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/RemoteAppClientExtensions.cs
twsouthwick: The null check is still there - this whole method was removed :)

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/RemoteAppServerExtensions.cs
twsouthwick: same - the method was removed. the null check is still in the existing method

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/RemoteAppClientExtensions.cs
mjrousos: Ah, right. 👍

Issue #195, opened 2022-09-08
Authentication redirect not working


pdunkerley:
I am having issues getting the redirect to the login page to work when accessing a Core page when I am not signed into the Framework app. I have a Core controller that has an [Authorize] attribute; when I am logged into the Framework app, the page displays just fine, but when I am not logged in, instead of being redirected to the login page, I get the following exception (this is actually from a logfile, but the same thing is also displayed in the browser):

```
2022-09-08 16:32:05 [13] ERROR DeveloperExceptionPageMiddleware.?():? - An unhandled exception has occurred while executing the request.
System.ArgumentException: An item with the same key has already been added. Key:                                 
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
   at System.Security.Claims.Claim..ctor(BinaryReader reader, ClaimsIdentity subject)
   at System.Security.Claims.ClaimsIdentity.CreateClaim(BinaryReader reader)
   at System.Security.Claims.ClaimsIdentity.Initialize(BinaryReader reader)
   at System.Security.Claims.ClaimsIdentity..ctor(BinaryReader reader)
   at System.Security.Claims.ClaimsIdentity.Initialize(BinaryReader reader)
   at System.Security.Claims.ClaimsIdentity..ctor(BinaryReader reader)
   at System.Security.Claims.ClaimsIdentity.Initialize(BinaryReader reader)
   at System.Security.Claims.ClaimsIdentity..ctor(BinaryReader reader)
   at System.Security.Claims.ClaimsIdentity.Initialize(BinaryReader reader)
   at System.Security.Claims.ClaimsIdentity..ctor(BinaryReader reader)
   at System.Security.Claims.ClaimsPrincipal.CreateClaimsIdentity(BinaryReader reader)
   at System.Security.Claims.ClaimsPrincipal..ctor(BinaryReader reader)
   at Microsoft.AspNetCore.SystemWebAdapters.Authentication.RemoteAppAuthenticationResultFactory.CreateRemoteAppAuthenticationResultAsync(HttpResponseMessage response, RemoteAppAuthenticationClientOptions options)
   at Microsoft.AspNetCore.SystemWebAdapters.Authentication.RemoteAppAuthenticationService.AuthenticateAsync(HttpRequest originalRequest, CancellationToken cancellationToken)
[...]

PR #209, opened 2022-09-26
Extract claims serialization logic into its own (overridable) service

mjrousos:
In order to make remote app authentication more flexible, this PR move claims serialization and deserialization logic into its own service which users can override with custom implementations, if needed. It also makes the `IAuthenticationResultFactory` interface public in case users need to customize processing of auth results even more broadly.

Related: #195


samples/MvcApp/Services/CustomClaimsSerializer.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Security.Claims;
+using Microsoft.AspNetCore.SystemWebAdapters.Authentication;
+
+namespace MvcApp.Services
+{
+    internal class CustomClaimsSerializer : IClaimsSerializer
+    {
+        public ClaimsPrincipal Deserialize(Stream responseContent)
+        {
+            throw new NotImplementedException();
```
twsouthwick, line 13: Why not have one serializer that shares code? That way it's clear that they serialize/deserialize to the same structure

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/Authentication/IAuthenticationResultFactory.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.AspNetCore.SystemWebAdapters.Authentication;
 /// Interface for creating authentication results from HTTP responses returned
 /// by a remote app in remote authentication scenarios.
 /// </summary>
-internal interface IAuthenticationResultFactory
+public interface IAuthenticationResultFactory
```
twsouthwick, line 13: We'll probably need to do an API review for this

src/Services/Authentication/IClaimsSerializer.cs
```diff
@@ -0,0 +1,11 @@
+using System.IO;
+using System.Security.Claims;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.SystemWebAdapters.Authentication;
+
+public interface IClaimsSerializer
+{
+    ClaimsPrincipal? Deserialize(Stream? responseContent);
+    void Serialize(ClaimsPrincipal? claimsPrincipal, Stream? outputStream);
```
twsouthwick, line 10: Why the nullable? Don't we want to only handle this if we are actually given a stream/principal? I understand the deserialize returning null potentially, but the others should probably be non-nullable

src/Services/Authentication/RemoteAppAuthenticationClaimsSerializer.cs
```diff
@@ -0,0 +1,36 @@
+using System;
+using System.IO;
+using System.Security.Claims;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.SystemWebAdapters.Authentication;
+
+internal class RemoteAppAuthenticationSerializer : IClaimsSerializer
```
twsouthwick, line 8: I don't see anything here that would be remote app specific. This seems like a general purpose claims principal serializer

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/Authentication/IAuthenticationResultFactory.cs
mjrousos: Agreed

src/Services/Authentication/RemoteAppAuthenticationClaimsSerializer.cs
mjrousos: Good point. Maybe `BinaryClaimsSerializer` instead?

src/Services/Authentication/IClaimsSerializer.cs
mjrousos: Good catch. I'll make the parameters non-nullable.

samples/MvcApp/Services/CustomClaimsSerializer.cs
mjrousos: Hm... I see what you're saying but it's not obvious where that shared source would live. Some options -

1. We could put it in ClassLibrary, but that project targets .NET Standard and the `IClaimsSerializer` interface is in the FrameworkServices/CoreServices packages which don't target .NET Standard. We could move `IClaimsSerializer` to Microsoft.AspNEtCore.SystemWebAdapters which does target .NET Standard, but I feel like the interface logically fits in the service-specific packages more than the adapters one.
2. We could have a shared source file that is included in both the MvcApp and MvcCoreApp sample projects. This is probably what I would do in a real-world scenario, but we don't currently have any shared source between those two sample projects and it's unclear to me if it's worth complicating our samples folder with an extra 'shared' folder or linking the file from one project to another for this case.

I could go either way on this one. I think the way it is in the PR now is ok, but if you prefer option 2 above that also seems perfectly reasonable and I'd be fine adding an MvcShared folder for the serializer class.

twsouthwick: This may be a good reason to have another package of `.Services`. The same issue will be hit with the key serialization. These are logically the same type, and we should probably do something so that they *are* the same type.

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/Authentication/IAuthenticationResultFactory.cs
twsouthwick: I'm not seeing a need to expose this for the serialization - am I missing something?

mjrousos: This is the interface whose implementation (`RemoteAppAuthenticationResultFactory`) uses the claims serialization. Making this public doesn't relate to the effort to make the serialization logic extensible directly, but it relates to the scenario of needing to customize how auth decisions are made in the ASP.NET Core app.

The advantage to this interface being public is that if, for some reason, a user's authentication solution was such that just changing how claims are serialized/deserialized wasn't enough to support getting identity from their ASP.NET app's authentication endpoint, they could implement this interface and have even broader control over how the response from the ASP.NET app is handled.

samples/MvcApp/Services/CustomClaimsSerializer.cs
mjrousos: Yea, if there was a `.Services` package, then the netstandard library used by the sample could depend on it and have the serializer implementation live there.

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/Authentication/IAuthenticationResultFactory.cs
mjrousos: After thinking about this more, I'm going to leave this as internal for now since there's not an immediate request for it to be public and it will be a simple one-line change to make it public in the future, if needed. I'll follow-up with API review folks on the new IClaimsSerializer interface.

adityamandaleeka:
Let's revisit this one post 1.0.

twsouthwick:
@mjrousos can we close this for now and reopen when we're ready to address it?

mjrousos:
Yep, that seems fine.

PR #213, opened 2022-09-29
Replace IRequestBodyPipeFeature if prebuffering

twsouthwick:
We currently replace the IHttpRequestFeature, but that gets the IRequestBodyPipeFeature.Reader out of sync if the stream is prebuffered. This change implements IRequestBodyPipeFeature on our custom feature and sets it as well.

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpRequestAdapterFeature.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpRequestAdapterFeature.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Diagnostics;
 using System.IO;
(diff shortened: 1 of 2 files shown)


twsouthwick:
@Tratcher are there guidelines about how to replace features that should be overridden as sets? I found this issue by trying to read a posted form and seeing that the form collection is created from a pipe, which would get a 0-length stream since it was using the original stream, but we had pre-buffered it. I had assumed that the request feature would be enough here 🤦

twsouthwick:
Just found this issue: https://github.com/dotnet/aspnetcore/issues/12620

Tratcher (approved):

Tratcher:
Yeah, we know this isn't ideal, but it hasn't been a big enough problem to address.

PR #229, opened 2022-10-14
Ensure multi-cookie headers are joined with ; instead of ,

### LINKED ISSUE #228: Remote authentication not working as expected
### Describe the bug

I have followed the incremental migration blog post series.

The setup correctly proxies to pages from the ASP.NET MVC application.  I am able to login and navigate the existing site as an authenticated user.  I can also see remote session variables are being requested successfully and accessible in ASP.NET Core MVC controller actions.

When I protect an endpoint in the ASP.NET Core MVC app with `[Authorize]` such as , `UserInfoController.Index` , I can see that a request to the endpoint results in a request to the remote authentication endpoint, `http://localhost:52187/systemweb-adapters/authenticate?original-url=%2FUserInfo`, as expected.  However, the authentication fails and I am redirected to the ASP.NET MVC login page.

By adding a breakpoint in `AccountController.Login`, I can verify that `HttpContext.User.Identity.IsAuthenticated` is `true` as I expect because I haven't logged out.

To debug the what was happening in `/systemweb-adapters/authenticate`, I had to copy several internal classes from Microsoft.AspNetCore.SystemWebAdapters, and configure them manually.

I found that `context.User.Identity.IsAuthenticated` in `RemoteAppAuthenticationHttpHandler.ProcessRequest()` is `false`.
This results in a `401` response, hence why I am being redirected back to the login page.
But then, in `AccountController.Login` the user is authenticated.

I don't know how to debug further than this.  I don't understand why `context.User.Identity.IsAuthenticated` in `RemoteAppAuthenticationHttpHandler.ProcessRequest()` is `false` and `HttpContext.User.Identity.IsAuthenticated` in `AccountController.Login` is `true`.  According to the blog post, the request should be authenticated and the user redirected back to the page from ASP.NET Core MVC app.

### To Reproduce

I am unsure how to reproduce this.

I have simply followed the examples in the incremental migration blog post series.

### Exceptions (if any)

None.

### Further technical details 

- Technologies and versions used (i.e. MVC/WebForms/etc): MVC, FormsAuthentication
- .NET Framework Version: 4.8

*ASP.NET Core Application*:
- Targeted .NET version: 6.0
- .NET SDK version: As per migration

mjrousos:
In cases of multi-headers, HttpClient concatenates the headers with `, `. In the case of multi-header *cookie* headers, though (which can happen because of https://github.com/dotnet/aspnetcore/issues/26461), [RFC 6265](https://datatracker.ietf.org/doc/html/rfc6265#section-4.2) specifies that multiple cookies must be concatenated with `; ` when writing the Cookie header.

HttpClient doesn't special case the Cookie header as it should, so this updates the RemoteAppAuthenticationService to special case cookie headers and concatenate them itself, if needed.

Fixes #228

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/Authentication/RemoteAppAuthenticationService.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/Authentication/RemoteAppAuthenticationService.cs
@@ -13,6 +13,7 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
+using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNetCore.SystemWebAdapters.Authentication;
 
@@ -97,7 +98,7 @@ internal partial class RemoteAppAuthenticationService : IRemoteAppAuthentication
     }
 
     // Add configured headers to the request, or all headers if none in particular are specified
-    private static void AddHeaders(IEnumerable<string> headersToForward, HttpRequest originalRequest, HttpRequestMessage authRequest)
+    internal static void AddHeaders(IEnumerable<string> headersToForward, HttpRequest originalRequest, HttpRequestMessage authRequest)
     {
         // Add x-forwarded headers so that the authenticate API will know which host the HTTP request was addressed to originally.
         // These headers are also used by result processors - to fix-up redirect responses, for example, to redirect back to the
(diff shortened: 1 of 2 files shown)


Tratcher (approved):

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/Authentication/RemoteAppAuthenticationService.cs
```diff
@@ -118,6 +119,16 @@ private static void AddHeaders(IEnumerable<string> headersToForward, HttpRequest
 
         foreach (var headerName in headerNames)
         {
+            // Workaround for an issue identified by https://github.com/dotnet/systemweb-adapters/issues/228.
+            // HttpClient wrongly uses comma (",") instead of semi-colon (";") as a separator for Cookie headers.
+            // To mitigate this, we concatenate them manually and put them back as a single header value.
+            // This workaround can be removed once we target .NET 7+ as HttpClient is fixed there.
```
Tratcher, line 125:
```suggestion
            // This workaround can be removed once we target .NET 7+ as Kestrel is fixed there.
```

PR #231, opened 2022-10-20
Use a singleton HttpClient for remote app communication

### LINKED ISSUE #221: RemoteAppClientOptions.BackchannelHandler should be a factory method
I've called `AddRemoteAppClient` like so:

    services.AddSystemWebAdapters()
        .AddRemoteAppClient(options =>
        {
            options.BackchannelHandler = new MyBackChannelHandler();
        }

Where _BackChannelHandler_ is basically just a HttpClientHandler with some logging. But a few times during testing I've got an `ObjectDisposedException: cannot access a disposed object. object name: 'SocketsHttpHandler'.` in my HttpClientHandler's SendAsync:

image

(sorry for the screenshot, didn't grab the plain text)

On investigation It would appear that HttpClientHandler should **be a new instance every time**, otherwise you eventually hit this ObjectDisposedException. See dotnet/runtime#27610 for another example.

Therefore I suggest that RemoteAppClientOptions.BackchannelHandler be changed so that it's something like a `Func<HttpMessageHandler>` instead, so that AddRemoteAppClient can call it to get a new instance each time. 

For example:

https://github.com/dotnet/systemweb-adapters/blob/9921217743fea54e602d95e22ac48f7ca915d261/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/RemoteAppClientExtensions.cs#L38

Change:

    if (options.BackchannelHandler is { } handler)
    {
        return handler;
    }

To something like:

    if (options.BackchannelHandlerFactory is { } factory)
    {
        return factory();
    }

mjrousos:
Updates remote app communication to use a singleton HttpClient instance (stored in RemoteAppClientOptions) rather than getting new clients from IHttpClientFactory for each request. This addresses an issue where HTTP handlers provided by users were being disposed and then reused by the factory previously.

Fixes #221

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/RemoteAppClientOptions.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/RemoteAppClientOptions.cs
@@ -60,7 +60,5 @@ public class RemoteAppClientOptions
     /// <summary>
     /// Gets or sets an <see cref="HttpClient"/> to use for making requests to the remote app.
     /// </summary>
-    public HttpClient BackchannelClient { get; set; }
-        // Set to default, because we'll ensure this is populated with post-configuration
-        = default!;
+    public HttpClient? BackchannelClient { get; set; }
 }
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/RemoteAppClientOptions.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/RemoteAppClientOptions.cs
@@ -53,7 +53,14 @@ public class RemoteAppClientOptions
     }
 
     /// <summary>
-    /// Gets or sets an <see cref="HttpMessageHandler"/> to use for making requests to the remote app.
+    /// Gets or sets an <see cref="HttpMessageHandler"/> to use for making requests to the remote app. Used if BackchannelClient is null.
     /// </summary>
     public HttpMessageHandler? BackchannelHandler { get; set; }
+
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/RemoteAppClientOptions.cs
```diff
@@ -53,7 +53,14 @@ public Uri RemoteAppUrl
...
-    /// Gets or sets an <see cref="HttpMessageHandler"/> to use for making requests to the remote app.
+    /// Gets or sets an <see cref="HttpMessageHandler"/> to use for making requests to the remote app. Used if BackchannelClient is null.
     /// </summary>
     public HttpMessageHandler? BackchannelHandler { get; set; }
+
+    /// <summary>
+    /// Gets or sets an <see cref="HttpClient"/> to use for making requests to the remote app.
+    /// </summary>
+    public HttpClient BackchannelClient { get; set; }
+        // Set to default, because we'll ensure this is populated with post-configuration
```
Tratcher, line 64: But it won't be populated for users interacting with this object before that. Nullable is more accurate. Your consuming code already handles the null case.

Tratcher (approved):

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/RemoteAppClientOptions.cs
mjrousos: Good point. I'll just make this nullable.

PR #234, opened 2022-10-25
Use default cancellation token on redirect

twsouthwick:
Redirect may cause the client to disconnect, but we still want to commit the session. Currently, this causes an exception to be thrown and the session will not be committed if a redirect has occurred.

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/SessionMiddleware.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/SessionMiddleware.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
+using System.Threading;
 using System.Threading.Tasks;
 using System.Web.SessionState;
 using Microsoft.AspNetCore.Http;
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/SessionMiddleware.cs
```diff
@@ -56,7 +57,11 @@ private async Task ManageStateAsync(HttpContextCore context, SessionAttribute me
         try
         {
             await _next(context);
-            await state.CommitAsync(context.RequestAborted);
+
+            // If the request is being redirected, the client may disconnect before we have a chance to commit, so we don't want to
+            // use HttpContext.RequestAborted as it will be disconnected
+            var token = context.GetAdapter().Response.IsRequestBeingRedirected ? CancellationToken.None : context.RequestAborted;
+            await state.CommitAsync(token);
```
Tratcher, line 64: Consider not using a cancellation token here at all, or at least, using one that's not associated with the client. They're not really involved at this stage. As long as there's a basic timeout to prevent this from starving you should be fine.

twsouthwick: That works. I've updated it to have a 1 min timeout

Tratcher (approved):

PR #235, opened 2022-10-25
Create a shared services library

twsouthwick:
With the current layering, the FrameworkServices and CoreServices libraries have some shared source. This is fine for internal implementations, but for public types, this causes two types to exist with the same name but different assemblies. This is not an issue in the main case (i.e. in an application) as they will never depend on each other.

However, the problem arises when a user wants to develop an implementation on a shared, public service that would then be consumed by both the framework services and core services. An example is the ISessionKeySerializer, that a user may want to customize for themselves, but the framework and core services have identical interfaces, but in different assemblies. These are logically the same service, and should therefore have the same identity (i.e. same assembly).


src/Services/Microsoft.AspNetCore.SystemWebAdapters.Services.csproj
```diff
@@ -0,0 +1,35 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
mjrousos, line 1: Nit-pick - maybe change the folder name to match the project name in order to match the pattern of other projects in this solution.

src/Services/BinaryWriterReaderExtensions.cs
```diff
@@ -1,6 +1,8 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+#if NETFRAMEWORK
```
mjrousos, line 4: Since this and HostString are excluded entirely from non-netfx targets, I wonder if it might be cleaner to just exclude them at the csproj level. Maybe move both files under a 'framework' folder and exclude that folder from compilation for TFMs other than net472.

Microsoft.AspNetCore.SystemWebAdapters.sln
```diff
@@ -59,6 +59,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Samples.MVCApp.Tests", "tes
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Samples.RemoteAuth.Forms.Tests", "test\Samples.RemoteAuth.Forms.Tests\Samples.RemoteAuth.Forms.Tests.csproj", "{7C994084-D425-487C-9266-413BB0C22695}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.SystemWebAdapters.Services", "src\Services\Microsoft.AspNetCore.SystemWebAdapters.Services.csproj", "{632E6195-4304-4C67-AABB-7CFC3F9086B6}"
```
mjrousos, line 62: I don't love the Microsoft.AspNetCore.SystemWebAdapters.Services name. It sounds like something that someone might use on its own (with .FrameworkServices and .CoreServices providing some TFM-specific helpers).

Maybe `Microsoft.AspNetCore.SystemWebAdapters.Common` or `Microsoft.AspNetCore.SystemWebAdapters.Abstractions` would better indicate that this is a shared support package that most users won't be depending on directly except for interfaces and things like that?

mjrousos (approved):
LGTM except that I'm not entirely sold on the name. A couple other comments inline.

src/Services/Microsoft.AspNetCore.SystemWebAdapters.Services.csproj
```diff
@@ -0,0 +1,35 @@
...
+    <RootNamespace>Microsoft.AspNetCore.SystemWebAdapters</RootNamespace>
+  </PropertyGroup>
+
+  <PropertyGroup Label="Package Information">
+    <Description>A collection of services to provide support for enabling configuration for Microsoft.AspNetCore.SystemWebAdapters.</Description>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <InternalsVisibleTo Include="Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices"/>
+    <InternalsVisibleTo Include="Microsoft.AspNetCore.SystemWebAdapters.CoreServices"/>
```
Tratcher, line 16: This tells me things aren't factored very cleanly. What's internal that the other projects need, and why doesn't it live in those assemblies instead?

twsouthwick: I've moved any implementation shared into a shared source pattern. Now, the only thing in the abstractions are the actual contracts

Microsoft.AspNetCore.SystemWebAdapters.sln
twsouthwick: I've renamed to abstractions based on feedback from the standup

src/Services/BinaryWriterReaderExtensions.cs
twsouthwick: moved back to the framework class only

twsouthwick:
Based on feedback, I've updated the package to be `Microsoft.AspNetCore.SystemWebAdapters.Abstractions` and only contain the public types that need to be shared. Turns out that the namespaces for the extension methods are not actually shared (i.e. `System.Web` vs `Microsoft.Extensions.DependencyInjection`), so those are still in the shared services.

This also allows the abstractions to be .NET Standard 2.0 so libraries that extend the types should be even simpler to produce.

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/SessionState/RemoteSession/RemoteAppSessionStateExtensions.cs
```diff
@@ -19,7 +19,6 @@ public static ISystemWebAdapterRemoteServerAppBuilder AddSessionServer(this ISys
 
         builder.Services.AddScoped<IHttpModule, RemoteSessionModule>();
         builder.Services.AddSingleton<ILockedSessionCache, InMemoryLockedSessions>();
-        builder.Services.AddSingleton<ISessionSerializer, BinarySessionSerializer>();
```
Tratcher, line 22: ?

Tratcher (approved):

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/SessionState/RemoteSession/RemoteAppSessionStateExtensions.cs
twsouthwick: Whoops - didn't realize I hadn't reverted that.

PR #238, opened 2022-11-11
Use HttpValueCollection as base class for custom collections

twsouthwick:
This enables .ToString() to function as similarly to .NET Framework as
possible. That's the only observable difference by adding this as the
base class.

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters/Internal/HttpValueCollection.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters/Internal/HttpValueCollection.cs
@@ -87,7 +87,7 @@ internal class HttpValueCollection : NameValueCollection
         }
     }
 
-    public override string ToString() => ToString(false);
+    public override string ToString() => ToString(true);
 
     private string ToString(bool urlencoded)
     {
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.SystemWebAdapters/Internal/HttpValueCollection.cs
```diff
@@ -87,7 +87,7 @@ internal void FillFromString(string s, bool urlencoded = false, Encoding? encodi
         }
     }
 
-    public override string ToString() => ToString(false);
+    public override string ToString() => ToString(true);
```
twsouthwick, line 90: not sure why this was false - on framework, this is true, so changing it back

src/Microsoft.AspNetCore.SystemWebAdapters/Internal/HttpValueCollection.cs
```diff
@@ -100,17 +100,17 @@ private string ToString(bool urlencoded)
 
         var s = new StringBuilder();
 
-        for (var i = 0; i < count; i++)
+        foreach (string k in this)
```
twsouthwick, line 103: this change is because we can't access by int, but rather by string

Tratcher (approved):

twsouthwick:
@Tratcher the build is failing with "Failed to request agent. Exception Image Build.Server.Amd64.VS2017.Open doesn't exist in pool NetCore-Public". Any ideas?

Tratcher:
No, that's a question for the first-responders channel.

PR #252, opened 2022-12-06
Enable CodeQL

adityamandaleeka:
Attempting to enable CodeQL in our CI pipeline.

@twsouthwick @dougbu @Tratcher


twsouthwick:
How do we verify that it works? I don't see anything in the builds about CodeQL - usually I've seen it as a separate check here

adityamandaleeka:
I set it to only run on internal builds, so I wouldn't expect to see anything new here.

adityamandaleeka:
@Tratcher I followed a general set of instructions that was sent to a bunch of teams (and therefore isn't at all tailored for our infrastructure/Arcade)... what @dougbu did in YARP is probably closer to what we should be doing here.

dougbu:
I also wanted to be as consistent as possible across the branches. Tasks are auto injected (when you let them be) _only_ in the repository's default branch.

Tratcher (approved):

PR #264, opened 2023-01-09
Add IEndpointRouteBuilder.MapBlazorPages extensions to explicitly map blazor pages

### LINKED ISSUE #53: Add support for migrating using Blazor
[Cocoon](https://github.com/VisualReCode/Cocoon) Offers support for migrating ASP.NET projects using Blazor. This is appealing for ASP.NET Web Form developers that are used to a component-based event-driven programming model.

Cocoon does some clever things to enable migration with Blazor using the strangler-fig pattern, like collecting the routable components in the app so that the fallback to the ASP.NET app happens correctly.

twsouthwick:
Currently, if you try to use Yarp with Blazor pages, it will not work as expected. Both of these currently expect to be the fallback. This adds in an extension method that will allow you to explicitly register blazor pages so they don't require a fallback anymore.

Fixes #53


src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/BlazorEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,87 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
```
twsouthwick, line 4: This is the interesting part - the rest is to add a sample

twsouthwick:
Blocked on API review here: https://github.com/dotnet/aspnetcore/issues/46170

gerneio:
@twsouthwick Just did some quick testing and unfortunately this new extension method `MapBlazorPages` throws an error if using Blazor pages with multiple `@page` directives (aka multiple routes). Blows up when calling `GetCustomAttribute`, however changing the code to loop over `GetCustomAttributes` seems to be working just fine.

Since the reason for this additional extension method is to help facilitate migration from older .NET Framework web projects (MVC/Web Forms) by using the strangler-fig pattern with YARP as a reverse-proxy (aka one-page-at-time sort of upgrade), that means for many of us we will want to make sure that our existing navigation links remain routable, while also taking advantage of the latest paradigm. For example, using web form pages will have the ASPX extension on every route by default. If I want to finally convert that page to Blazor, then I'm going to want to add multiple routes to my razor file so that the ASPX route still routes to the new stuff, but my new "default" route excludes the ASPX extension if possible. Here's an example of what that razor page would look like, although I'm sure there is probably a more elegant way to handle this.

```
@page "/page1"
@page "/page1.aspx"
@inject NavigationManager NavigationManager

<PageTitle>Converted Page 1</PageTitle>

@code {
    protected override void OnInitialized()
    {
        var path = NavigationManager.ToBaseRelativePath(NavigationManager.Uri);

        if (path.Contains("aspx"))
            NavigationManager.NavigateTo(path.Replace(".aspx", ""), replace: true);
    }
}
```

Hope this makes sense.

twsouthwick:
@gerneio thanks for the feedback. We're going to publish this as a doc but it will be available in ASP.NET Core 8 out of the box.

I'm a little confused by the rewriting you're describing - is this a feature request or feedback on how your using it? We aren't really opinionated on what the final routes end up being, but I would imagine that main constraint would be that you want existing links to continue working.

gerneio:
@twsouthwick I was just trying to show why someone would want multiple routes on their razor pages when migrating (for example) their web forms projects using the one-page-at-a-time approach (aka strangler-fig).

TLDR, this new extension method should support pages with multiple routes, rather than throwing an exception.

Btw, on those docs y'all are fixing to create, a mention of how to tackle the re-routing (i.e "/page.aspx" to "/page" IF the route exist in the new project) would be greatly welcomed. I managed to get something workable for my solution, but I have a feeling there are better approaches. 

Thanks!

twsouthwick:
cool - thanks for trying it out @gerneio . I'm going to close this as we're going to put it in docs instead.

gerneio:
@twsouthwick sounds good to me. Btw, what can I subscribe to so that I am notified when that doc has been completed?

PR #271, opened 2023-01-23
Add skippable endpoint extensions

twsouthwick:
A common setup with the adapters is to pair it with YARP. In this setup, routes are first served from the core application if available, and then proxied to the backend if not. However, there may be times where people will want to control whether an endpoint is actually served from the core app or just proxied to the backend. A common example of this is deployment testing while migrating.

This change enables this control by:

1) Implement a new interface  ISkippableEndpointSelector that says if an endpoint should be skipped
2) Add metadata to endpoints that may be skipped by calling `.EnableSkipping()` when mapping a route


CZEMacLeod:
Ooo - I like this. Would it be possible to access session state during the skip check, or would it not be loaded at that point? I think I would like to be able to toggle it on or off, either per person or per session, rather than via query parameters (although that is also good for many scenarios).

twsouthwick:
> Would it be possible to access session state during the skip check

This hooks into the routing, so is invoked in the `app.UseRouting()` middleware. Things like authentication/authorization/system web adapters run after this, so that won't be available.

> I think I would like to be able to toggle it on or off, either per person or per session, rather than via query parameter

The query example is more of a simple example for the sample. The scenarios I've seen users needing would be to integrate it into their feature flag system.

Would those constraints be workable for you?

CZEMacLeod:
@twsouthwick Usable as is - a feature flag is fine, but it would be much better to be able to do an action in one, then flip the setting and try again on the other to check the same functionality. Especially without losing session and client state. 
Again for A/B testing or client sign-off where you could allow some users to try the new code while everyone else used the original app underneath.
I'm not sure if the checks are executed on each request, or when building the routing table. I do much more ASP.Net 4.x coding than dotnet 6+.
Is there a way to 'abort' the request and fallback inside a controller - when you have all the additional information to hand? It would mean all requests hit the controller and then only some execute there... I don't know if that could work with the way that aspnetcore does forward only reading etc.

HaoK (commented):
Looks good, is there an easy way to add a test for these new extensions?

twsouthwick:
@HaoK added a test

twsouthwick:
@CZEMacLeod 

> I'm not sure if the checks are executed on each request, or when building the routing table. I do much more ASP.Net 4.x coding than dotnet 6+.

This check does get executed on each request so that you can decide on a per-request basis.

> Is there a way to 'abort' the request and fallback inside a controller

What do you mean by "abort"? Once an endpoint is executing, it will go to completion. A controller is just an endpoint so you have to decide which is run before you get to the "UseEndpoints()" middleware.

HaoK (approved):
Great, looks good once you fix the sealed warnings :)

twsouthwick:
I'm going to close this for now as there are some competing designs we need to agree on first: https://github.com/dotnet/systemweb-adapters/issues/272

PR #276, opened 2023-02-03
Full support for CacheDependency and accessors

### LINKED ISSUE #26: Consider supporting cache dependencies
Currently the Cache implementation exposes `CacheDepenendency` but cannot be constructed or used at the moment. This was used in framework to tie a cache entry to be updated/removed when some dependency changed. The default implementation watched for changes in files/directories. See https://docs.microsoft.com/en-us/dotnet/api/system.web.caching.cachedependency for details

CZEMacLeod:
**Full support for CacheDependency and accessors.**
Complete the implementation of `System.Web.Caching.CacheDependency` and provide all the accessors from System.Web implementation.

**Description**
Implement all constructors and features of `System.Web.Caching.CacheDependency`
Add extensions to allow handling various underlying `ChangeMonitor` implementations to surface the `UtcLastModified` property.
Implements `CacheDependencyChangeMonitor` to allow a `CacheDependency` to rely on another item in the Cache.
Adds `Cache` accessor to `HttpContext`, `HttpContextBase`, `HttpContextWrapper`, and `HttpRuntime`
Modifies `HttpRuntimeFactory` to allow getting the Cache implementation
Updates tests to support new `IHttpRuntime` interface member.

Addresses #26

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpRuntimeFactory.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpRuntimeFactory.cs
@@ -2,42 +2,60 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
+using System.Web.Caching;
 using Microsoft.AspNetCore.SystemWebAdapters;
 
 namespace Microsoft.Extensions.DependencyInjection;
 
 internal static class HttpRuntimeFactory
 {
-    public static IHttpRuntime Create()
+    public static IHttpRuntime Create(IServiceProvider serviceProvider)
     {
         if (NativeMethods.IsAspNetCoreModuleLoaded())
         {
             var config = NativeMethods.HttpGetApplicationProperties();
-
-            return new IISHttpRuntime(config);
+            return new IISHttpRuntime(config, serviceProvider);
         }
 
-        return new DefaultHttpRuntime();
+        return new DefaultHttpRuntime(serviceProvider);
     }
 
-    internal class DefaultHttpRuntime : IHttpRuntime
+    internal abstract class BaseHttpRuntime : IHttpRuntime
     {
-        public string AppDomainAppVirtualPath => "/";
+        private readonly IServiceProvider serviceProvider;
+        private Cache? cache;
 
-        public string AppDomainAppPath => AppContext.BaseDirectory;
+        protected BaseHttpRuntime(IServiceProvider serviceProvider)
+        {
+            this.serviceProvider = serviceProvider;
+        }
+        public abstract string AppDomainAppVirtualPath { get; }
+        public abstract string AppDomainAppPath { get; }
+        public Cache Cache => cache ??= serviceProvider.GetRequiredService<Cache>();
     }
 
-    internal class IISHttpRuntime : IHttpRuntime
+    internal class DefaultHttpRuntime : BaseHttpRuntime
+    {
+        public DefaultHttpRuntime(IServiceProvider sp) : base(sp)
+        {
+        }
+
+        public override string AppDomainAppVirtualPath => "/";
+
+        public override string AppDomainAppPath => AppContext.BaseDirectory;
+    }
+
+    internal class IISHttpRuntime : BaseHttpRuntime
     {
         private readonly NativeMethods.IISConfigurationData _config;
 
-        public IISHttpRuntime(NativeMethods.IISConfigurationData config)
+        public IISHttpRuntime(NativeMethods.IISConfigurationData config, IServiceProvider sp) : base(sp)
         {
             _config = config;
         }
 
-        public string AppDomainAppVirtualPath => _config.pwzVirtualApplicationPath;
+        public override string AppDomainAppVirtualPath => _config.pwzVirtualApplicationPath;
 
-        public string AppDomainAppPath => _config.pwzFullApplicationPath;
+        public override string AppDomainAppPath => _config.pwzFullApplicationPath;
     }
 }
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpRuntimeFactory.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpRuntimeFactory.cs
@@ -20,32 +20,29 @@ internal static class HttpRuntimeFactory
         return new DefaultHttpRuntime(serviceProvider);
     }
 
-    internal abstract class BaseHttpRuntime : IHttpRuntime
+    internal abstract class BaseHttpRuntime 
     {
         private readonly IServiceProvider serviceProvider;
-        private Cache? cache;
 
         protected BaseHttpRuntime(IServiceProvider serviceProvider)
         {
             this.serviceProvider = serviceProvider;
         }
-        public abstract string AppDomainAppVirtualPath { get; }
-        public abstract string AppDomainAppPath { get; }
-        public Cache Cache => cache ??= serviceProvider.GetRequiredService<Cache>();
+        public Cache Cache => serviceProvider.GetRequiredService<Cache>();
     }
 
-    internal class DefaultHttpRuntime : BaseHttpRuntime
+    internal class DefaultHttpRuntime : BaseHttpRuntime, IHttpRuntime
     {
         public DefaultHttpRuntime(IServiceProvider sp) : base(sp)
         {
         }
 
-        public override string AppDomainAppVirtualPath => "/";
+        public string AppDomainAppVirtualPath => "/";
 
-        public override string AppDomainAppPath => AppContext.BaseDirectory;
+        public string AppDomainAppPath => AppContext.BaseDirectory;
     }
 
-    internal class IISHttpRuntime : BaseHttpRuntime
+    internal class IISHttpRuntime : BaseHttpRuntime, IHttpRuntime
     {
         private readonly NativeMethods.IISConfigurationData _config;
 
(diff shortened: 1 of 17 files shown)


Tratcher:
Test failure:
```
    Microsoft.AspNetCore.SystemWebAdapters.HttpContextTests.CacheFromServices [FAIL]
      System.InvalidOperationException : HttpRuntime is not available in the current environment
      Stack Trace:
        /_/src/Microsoft.AspNetCore.SystemWebAdapters/HttpRuntime.cs(14,0): at System.Web.HttpRuntime.get_Current()
        /_/src/Microsoft.AspNetCore.SystemWebAdapters/HttpRuntime.cs(25,0): at System.Web.HttpRuntime.get_Cache()
        /_/src/Microsoft.AspNetCore.SystemWebAdapters/HttpContext.cs(45,0): at System.Web.HttpContext.get_Cache()
        /_/test/Microsoft.AspNetCore.SystemWebAdapters.Tests/HttpContextTests.cs(152,0): at Microsoft.AspNetCore.SystemWebAdapters.HttpContextTests.CacheFromServices()
```

Tratcher (approved):
What's the thread safety model here? I'm a little surprised not to see any locking anywhere.

CZEMacLeod:
> What's the thread safety model here? I'm a little surprised not to see any locking anywhere.

The Cache implementation is a wrapper around a `System.Runtime.Caching.ObjectCache` implementation (`System.Runtime.Caching.MemoryCache` by default) and all the locking etc. logic would be implemented there.

I don't think any of the code I've added actually requires any specific locking logic, although maybe I've missed a scenario where it would be applicable?

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpRuntimeFactory.cs
```diff
@@ -2,42 +2,60 @@
...
+    internal abstract class BaseHttpRuntime : IHttpRuntime
+    {
+        private readonly IServiceProvider serviceProvider;
+        private Cache? cache;
+
+        protected BaseHttpRuntime(IServiceProvider serviceProvider)
+        {
+            this.serviceProvider = serviceProvider;
+        }
+        public abstract string AppDomainAppVirtualPath { get; }
```
twsouthwick, line 32: This type itself doesn't need to implement IHttpRuntime, just the stuff that is shared and requires the service provider. The IHttpRuntime can be just added to the DefaultHttpRuntime/IIS

src/Microsoft.AspNetCore.SystemWebAdapters/HttpContext.cs
```diff
@@ -41,7 +41,9 @@ internal HttpContext(HttpContextCore context)
 
     public HttpServerUtility Server => _server ??= new(_context);
 
-    public Cache Cache => _context.RequestServices.GetRequiredService<Cache>();
+#pragma warning disable CA1822 // Mark members as static
+    public Cache Cache => HttpRuntime.Cache;
```
twsouthwick, line 45: Please still retrieve it through the service provider - it keeps testing easier

src/Microsoft.AspNetCore.SystemWebAdapters/HttpContextWrapper.cs
```diff
@@ -72,5 +51,7 @@ public override IPrincipal User
             get => _context.User;
             set => _context.User = value;
         }
+
+        public override Cache Cache => HttpRuntime.Cache;
```
twsouthwick, line 55:
```suggestion
        public override Cache Cache => _context.Cache;
```

test/Microsoft.AspNetCore.SystemWebAdapters.CoreServices.Tests/CacheTests.cs
```diff
@@ -0,0 +1,82 @@
...
+        public void CacheFromHttpContext()
+        {
+            // Arrange
+            var cache = new Cache();
+
+            var serviceProvider = new Mock<IServiceProvider>();
+            serviceProvider.Setup(s => s.GetService(typeof(Cache))).Returns(cache);
+
+            var httpRuntime = Microsoft.Extensions.DependencyInjection.HttpRuntimeFactory.Create(serviceProvider.Object);
+            HttpRuntime.Current = httpRuntime;
```
twsouthwick, line 31: Any usage of the HttpRuntime outside of statics themselves should be done through DI so we don't need to set this in tests.

test/Microsoft.AspNetCore.SystemWebAdapters.CoreServices.Tests/CacheTests.cs
```diff
@@ -0,0 +1,82 @@
...
+        public void CacheFromHttpContextWrapper()
+        {
+            // Arrange
+            var cache = new Cache();
+
+            var serviceProvider = new Mock<IServiceProvider>();
+            serviceProvider.Setup(s => s.GetService(typeof(Cache))).Returns(cache);
+
+            var httpRuntime = Microsoft.Extensions.DependencyInjection.HttpRuntimeFactory.Create(serviceProvider.Object);
+            HttpRuntime.Current = httpRuntime;
```
twsouthwick, line 52: Same as above

test/Microsoft.AspNetCore.SystemWebAdapters.CoreServices.Tests/CacheTests.cs
```diff
@@ -0,0 +1,82 @@
...
+        public void CacheFromHttpRuntime()
+        {
+            // Arrange
+            var cache = new Cache();
+
+            var serviceProvider = new Mock<IServiceProvider>();
+            serviceProvider.Setup(s => s.GetService(typeof(Cache))).Returns(cache);
+
+            var httpRuntime = Microsoft.Extensions.DependencyInjection.HttpRuntimeFactory.Create(serviceProvider.Object);
+            HttpRuntime.Current = httpRuntime;
```
twsouthwick, line 75: same as above

test/Microsoft.AspNetCore.SystemWebAdapters.Tests/HttpContextTests.cs
```diff
@@ -134,26 +134,7 @@ public void ItemsNotWrappedIfAlreadyImplementsIDictionary()
             Assert.Same(items, result);
         }
 
-        [Fact]
```
twsouthwick, line 137: let's keep this as we want it to come from the services. We can expose it through a static property, but internally we shouldn't rely on it that way

twsouthwick (changes requested):
Looks great! My main concern is around the use of the static HttpRuntime. I'm fine exposing APIs that were there in ASP.NET Framework, but internally, we shouldn't be using those statics for things but access it through the DI container.

CZEMacLeod:
> We shouldn't be using those statics for things but access it through the DI container.

It think it was a combination of copying from the original design in net4 and thoughts to performance, as if the hit DI _every time_ we need access to the cache, rather than using a static property that is initialized once from the singleton, it might hurt the performance. 
This would happen e.g. when trying to get an item from the cache, not finding it, then adding it.
https://learn.microsoft.com/en-us/previous-versions/aspnet/xhy3h9f9(v=vs.100)
```cs
string cachedString;
cachedString = (string)HttpRuntime.Cache["CacheItem"];
if (cachedString == null)
{
  cachedString = "Hello, World.";
  HttpRuntime.Cache.Insert("CacheItem", cachedString);
}
```
Where `HttpRuntime` could also be `HttpContext`/`HttpContextBase`.

I have removed this from `BaseHttpRuntime`, and now all the accessor methods always hit DI. I guess if it is an issue, then the cache can be stored in a local variable in code. https://github.com/dotnet/systemweb-adapters/commit/685ce4642311d325cdaac71721f4d2a12bc63d9d#r100106939

test/Microsoft.AspNetCore.SystemWebAdapters.CoreServices.Tests/CacheTests.cs
```diff
@@ -0,0 +1,95 @@
...
+
+            var coreContext = new Mock<HttpContextCore>();
+            coreContext.Setup(c => c.RequestServices).Returns(serviceProvider.Object);
+
+            var context = new HttpContext(coreContext.Object);
+            
+            // Act
+            var result = context.Cache;
+
+            // Assert
```
twsouthwick, line 33:
```suggestion
            // Assert
            Assert.Same(cache, result);
```

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpRuntimeFactory.cs
```diff
@@ -2,36 +2,51 @@
...
+
+    internal abstract class BaseHttpRuntime 
+    {
+        private readonly IServiceProvider serviceProvider;
+
+        protected BaseHttpRuntime(IServiceProvider serviceProvider)
+        {
+            this.serviceProvider = serviceProvider;
+        }
+        public Cache Cache => serviceProvider.GetRequiredService<Cache>();
```
twsouthwick, line 31:
```suggestion
        protected BaseHttpRuntime(IServiceProvider serviceProvider)
        {
            Cache = serviceProvider.GetRequiredService<Cache>();
        }

        public Cache Cache { get; }
```

twsouthwick: I'm fine with caching the cache ( :) ), but it's still being accessed via the DI system initially

twsouthwick (approved):

PR #286, opened 2023-02-09
Add HttpRequest.Flush[Async]

### LINKED ISSUE #279: Add missing API: HttpResponse.Flush
## Summary
 When I try the IHttpModule and HttpApplication in the adapter, I found that HttpResponse.Flush is used in my module(in Sharepoint Online) but it is not supported.

## Motivation and goals
HttpResponse.Flush is used in my module(in Sharepoint Online) but it is not supported. I wonder if there is plan to support the API.

## In scope
HttpResponse.Flush

## Usage example:
```
Response.StatusCode = 500;
Response.Clear();
Response.Flush();
Response.End();
```

twsouthwick:
Fixes #279

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters/HttpResponse.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters/HttpResponse.cs
@@ -5,6 +5,7 @@ using System.Collections.Specialized;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Text;
+using System.Threading.Tasks;
 using Microsoft.AspNetCore.Diagnostics;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.SystemWebAdapters/HttpResponse.cs
```diff
@@ -252,6 +253,10 @@ private string ResolvePath(string url)
 
         public void SetCookie(HttpCookie cookie) => Cookies.Set(cookie);
 
+        public void Flush() => _response.CompleteAsync().GetAwaiter().GetResult();
+
+        public Task FlushAsync() => _response.CompleteAsync();
```
Tratcher, line 258: CompleteAsync is more equivalent to End(). Response.Body.Flush/Async() is a closer match.

twsouthwick: Good call - updated

Tratcher (approved):

PR #289, opened 2023-02-14
Add HttpSessionState.Mode

### LINKED ISSUE #281: Add missing API: HttpSessionState.Mode
## Summary
 When I try the IHttpModule and HttpApplication in the adapter, I found that HttpSessionState.Mode is used in my module(in Sharepoint Online) but it is not supported.

## Motivation and goals
HttpSessionState.Mode is used in my module(in Sharepoint Online) but it is not supported. I wonder if there is plan to support the API.

## In scope
HttpSessionState.Mode

## Usage example:
```
context.Session.Mode != SessionStateMode.Off
```

twsouthwick:
This change exposes the API and sets it to `Custom`. Everything else has a meaning and implies certain behavior. This value is the only one that makes sense for us to declare. If needed later we can make it customizeable.

Fixes #281

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/SessionState/Wrapped/AspNetCoreSessionManager.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/SessionState/Wrapped/AspNetCoreSessionManager.cs
@@ -2,7 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Threading.Tasks;
-using System.Web.SessionState;
 using Microsoft.AspNetCore.SystemWebAdapters.SessionState.Serialization;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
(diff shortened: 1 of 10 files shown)


twsouthwick:
@Tratcher We had talked about creating an option to enable this, but this ties it to the ISessionManager implementation with a default interface method. Thoughts?

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/SessionState/Wrapped/AspNetCoreSessionManager.cs
```diff
@@ -23,4 +24,6 @@ public AspNetCoreSessionManager(ISessionKeySerializer serializer, ILoggerFactory
 
     public Task<ISessionState> CreateAsync(HttpContextCore context, SessionAttribute metadata)
         => Task.FromResult<ISessionState>(new AspNetCoreSessionState(context.Session, _serializer, _loggerFactory, metadata.IsReadOnly, _options.Value.ThrowOnUnknownSessionKey));
+
+    SessionStateMode ISessionManager.Mode => SessionStateMode.InProc;
```
Tratcher, line 28: This is misleading, you can't tell if ISession is storing data in-proc or remote. Will that matter? 🤷

Tratcher (approved):

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/SessionState/Wrapped/AspNetCoreSessionManager.cs
twsouthwick: Good call. After looking into it, the only thing that really makes sense for us to decalre ourselves as is Custom - everything else has a meaning.

twsouthwick:
After chatting with @mjrousos we decided that the value Customize was the only one that made sense for now.

Tratcher (approved):
Hence the original idea to make it configurable so it would say whatever they needed it to say, but this is adequate for now.

PR #290, opened 2023-02-15
Make browser APIs internal

twsouthwick:
These probably make sense to be public, but let's do that after they've been reviewed.


Tratcher:
Is there any hurry on this, is there a release coming up? Better to schedule the review than to change this and then revert it.

twsouthwick:
I guess not - I just didn't want to forget about it.

twsouthwick:
@Tratcher we decided to get a v1.1 out in the next week to unblock a partner, so we should merge this in and address it for v1.2

Tratcher (approved):

PR #294, opened 2023-02-17
Always add adapter features into context

twsouthwick:
Currently, the adapters replace a few features if endpoint metadata is configured for this to happen. However, this was intended to gate "expensive" operations, such as prebuffering the stream to enable InputStream. However, this now prevents APIs that are not "expensive" to be gated as well but are implemented in the feature as well. Some examples of this are `HttpRequest.End()` that is commonly used - it makes sense to implement it at the features, but shouldn't require buffering the response to turn it on.

This change will still keep the gating in place for the "expensive" operations, but enable APIs that are provided by the features to be available if possible without the endpoint metadata. It also adds some E2E tests using TestHost to verify these APIs perform as expected in the pipeline.


src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/RegisterAdapterFeaturesMiddleware.cs
```diff
@@ -0,0 +1,63 @@
...
+        RegisterRequestFeatures(context);
+        var buffering = RegisterResponseFeatures(context);
+
+        try
+        {
+            await _next(context);
+        }
+        finally
+        {
+            await buffering.FlushAsync();
```
Tratcher, line 31: Revert to the original features.

Tratcher (approved):

PR #295, opened 2023-02-17
Refactor HttpRuntime usage to minimize statics usage

twsouthwick:
In ASP.NET Framework, HttpRuntime was used as a static holder for the runtime information. We expose this type in the adapters, but it should not be accessed directly internally due to its static nature. This currently causes sporadic test failures. Rather, internally should access the IHttpRuntime if this information is needed.

This change includes:

- Changes UrlPath to be a class that takes in the IHttpRuntime
- Moves the logic of VirtualPathUtility into a VirtualPathUtilityImpl that is an instance class
- VirtualPathUtility uses a cached instance of VirtualPathUtilityImpl when needed
- Current usage of VirtualPathUtility in the adapters is replaced with the VirtualPathUtilityImpl
- Tests are updated to no longer require setting HttpRuntime.Current

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters/Caching/CacheDependency.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters/Caching/CacheDependency.cs
@@ -10,7 +10,6 @@ namespace System.Web.Caching;
 public class CacheDependency : IDisposable
 {
     private readonly List<ChangeMonitor> changeMonitors = new();
-    private bool hasChanged;
     private bool disposedValue;
     private DateTime utcLastModified;
     private Action<object, EventArgs>? dependencyChangedAction;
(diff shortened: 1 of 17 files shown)


Tratcher (approved):

CZEMacLeod:
Just a couple of quick notes - this should probably mention that it includes a heap of changes to add copyright information to some files not associated with HttpRuntime or the VPU stuff.
Also changes Cache to match code style (no single line ifs, and removing backing fields).
I think the ref file should _not_ have `System.Runtime.CompilerServices.CompilerGeneratedAttribute` in it, as this does not match the native dll, and is an implementation detail.
That attribute should maybe be added to the excluded attributes file instead. (src/Microsoft.AspNetCore.SystemWebAdapters/Generated/ExcludedAttributes.txt)

twsouthwick:
Created #298 to remove the attribute. Good call.

All the code is inspired or taken from referencesource so not sure we need specific call outs for that. I'll follow up on that.

PR #298, opened 2023-02-21
Remove CompilerGeneratedAttribute from generated .NET Standard assembly

twsouthwick:
This is an implementation detail that we don't need to leak into the .NET Standard assembly


CZEMacLeod:
Is there a reason this was closed without merging?

PR #314, opened 2023-03-27
Add IHttpModule and pipeline infrastructure to emulate it

twsouthwick:
This change adds support for `HttpApplication` and `IHttpModule` and is emulated as best as possible on the ASP.NET Core pipeline. This is not tied to IIS and will work on Kestrel or any other host by using middleware to invoke the expected events at the times that best approximate the timing from ASP.NET Core. An attempt has been made to get the events to fire at the appropriate time, but because of the substantial difference between ASP.NET and ASP.NET Core there may still be unexpected behavior.

In order to register either an `HttpApplication` or `IHttpModule` instance, use the following pattern:

```csharp
using System.Web;
using ModulesLibrary;

var builder = WebApplication.CreateBuilder(args);

    // Non-generic version available if no custom HttpApplication is needed
    .AddHttpApplication<MyApp>(options =>
    {
        // Size of pool for HttpApplication instances. Should be what the expected concurrent requests will be
        options.PoolSize = 10;
        // Register a module by name
        options.RegisterModule<MyModule>("Module");
    });

var app = builder.Build();

app.UseSystemWebAdapters();

app.Run();

class MyApp : HttpApplication
{
  protected void Application_Start()
  {
    ...
  }
  protected void Session_Start()
  {
    ...
  }
  protected void Begin_Request()
  {
    ...
  }
  ...
}

class MyModule : IHttpModule
{
  public void Init(HttpApplication app)
  {
    ...
  }
  
  public void Dispose()
  {
  }
}
```

The normal `.UseSystemWebAdapters()` middleware builder will enable majority of the events. However, the authentication and authorization events require two additional middleware calls in order to enable them if you want the events to fire in the expected order. If they are omitted, they will be called at the point `UseSystemWebAdapters()` is added.

```diff
app.UseRouting();
app.UseAuthentication();
+ app.UseRaiseAuthenticationEvents();
app.UseAuthorization();
+ app.UseRaiseAuthorizationEvents();
app.UseSystemWebAdapters();
```

## When should this be used?

> Most of the time, this should not be used. Prefer direct ASP.NET Core middleware if possible.

This is intended mostly for scenarios where a module needs to be run on ASP.NET Core but is unable to be migrated easily. Ideally, the code in a module should be restructured to be used as middleware. This is especially recommended when only a single or few events are used; those can usually be migrated in a straightfoward way.

However, if a module has many thousands of line of code and many events being used (the initial driver of this feature), this can provide a stepping stone to migrating that functionality to ASP.NET Core.

## Emulated Events

> For details on how this worked in .NET Framework, see the [official documentation](https://learn.microsoft.com/en-us/dotnet/api/system.web.httpapplication)

The IIS event pipeline that is expected by `IHttpModule` and `HttpApplication` is emulated using middleware by the adapters. As part of this, it will add additional middleware that will invoke the events. This is done via a feature that is inserted early on in the adapter pipeline.  This exposes the `HttpApplication` for the request, as well as the ability to raise events on it.

Events have a prescribed order which is replicated with these emulated events. However, because the rest of the ASP.NET Core pipeline is unaware of these events and so some of the state of the request may not be exactly replicated.

A common pattern is to be able to call `HttpRequest.End()` or `HttpApplication.CompleteRequest()`. Both of these are supported, as well as continuing to raise the events that are raised in IIS with this (including `EndRequest` and the logging events).

> Note: In the cases in which no modules or `HttpApplication` type is registered, the emulated pipeline is not added to the middleware chain.

## HttpApplication lifetime

On ASP.NET Framework, each request would get an individual `HttpApplication` instance. This object contains the following information:

- Event callbacks registered either on the `HttpApplication` type itself or on registered modules
- Any state contained in the `HttpApplication` instance or its registered modules

In order to support this, one of the first middlewares invoked will retrieve an instance of `HttpApplication`. This uses a `PooledObjectPolicy<HttpApplication>` that will create an instance of the application's `HttpApplication` type and register all modules on it. When the request is exiting that middleware, it will return the `HttpApplication` instance to the pool which will also remove the `HttpContext` instance assigned to it.

This can potentially create a number of instances of `HttpApplication` that are only used a limited number of times. The pool can be controlled by customizing the `HttpApplicationOptions.PoolSize` option or providing a custom implementation of `ObjectPool<HttpApplication>` that can use the `PooledObjectPolicy<HttpApplication>` provided to override the pooling behavior.


src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/HttpApplicationMiddleware.cs
```diff
@@ -0,0 +1,61 @@
...
+                await context.Features.GetRequired<IHttpApplicationFeature>().RaiseEventAsync(ApplicationEvent.RequestCompleted);
+            }
+        }
+        finally
+        {
+            _pool.Return(app);
+        }
+    }
+
+    private static DelegateDisposable SetRequiredFeatures(HttpContextCore context, HttpApplication application)
```
davidfowl, line 47: This feels unnecessary since its only used in a single place.

twsouthwick: Initially the InvokeAsync was doing a lot more so I think it made sense then, but I'm fine with inlining it. Done

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/HttpApplicationExtensions.cs
```diff
@@ -0,0 +1,157 @@
...
+                {
+                    await appFeature.RaiseEventAsync(@event);
+
+                    if (endFeature.IsEnded)
+                    {
+                        return;
+                    }
+                }
+
+                await next(ctx);
```
Tratcher, line 91: should the postEvents still run if an exception is thrown?

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/HttpApplicationExtensions.cs
```diff
@@ -0,0 +1,157 @@
...
+                        return;
+                    }
+
+                    await appFeature.RaiseEventAsync(@event);
+                }
+            });
+        }
+    }
+
+    public static IApplicationBuilder UseRaiseAuthenticationEvents(this IApplicationBuilder app)
```
Tratcher, line 106:
```suggestion
    public static IApplicationBuilder UseAuthenticationEvents(this IApplicationBuilder app)
```

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/HttpApplicationMiddleware.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            context.Features.Set(app);
+            app.Context = context;
+
+            context.Features.GetRequired<IHttpResponseBufferingFeature>().EnableBuffering(BufferResponseStreamAttribute.DefaultMemoryThreshold, default);
+
+            await _next(context);
+
+            await context.Features.GetRequired<IHttpResponseEndFeature>().EndAsync();
+            await context.Features.GetRequired<IHttpApplicationFeature>().RaiseEventAsync(ApplicationEvent.RequestCompleted);
```
Tratcher, line 42: What about exceptions here?

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/RequestHttpApplicationFeature.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        Application.InvokeEvent(@event);
+
+        return ValueTask.CompletedTask;
+    }
+
+    async Task IHttpResponseEndFeature.EndAsync()
+    {
+        // Prevent subsequent calls from going through this pathway again
+        _context.Features.Set(NoopEndFeature.Instance);
```
Tratcher, line 71: This only works if the next time they fetch the feature again, not if they're already holding onto an existing one? Would it be better to track state in this instance and reset it later if you're pooling these.

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/SessionMiddleware.cs
```diff
@@ -34,13 +34,28 @@ public SessionMiddleware(RequestDelegate next, ILogger<SessionMiddleware> logger
...
+
+    private Task NoSessionAsync(HttpContextCore context)
+        => context.Features.Get<IHttpApplicationFeature>() is { } events ? RaiseEventsNoSessionAsync(context, events) : _next(context);
+
+    private async Task RaiseEventsNoSessionAsync(HttpContextCore context, IHttpApplicationFeature appFeature)
+    {
+        await appFeature.RaiseEventAsync(ApplicationEvent.AcquireRequestState);
+        await appFeature.RaiseEventAsync(ApplicationEvent.PostAcquireRequestState);
+
+        await _next(context);
```
Tratcher, line 47: Exceptions?

Tratcher (approved):

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/HttpApplicationMiddleware.cs
```diff
@@ -0,0 +1,61 @@
...
+
+    public HttpApplicationMiddleware(RequestDelegate next, ObjectPool<HttpApplication> pool)
+    {
+        _next = next;
+        _pool = pool;
+    }
+
+    public async Task InvokeAsync(HttpContextCore context)
+    {
+        var app = _pool.Get();
```
davidfowl, line 24: Do we need to preserve the pooling behavior at all? Can't we just have an allocation per request here?

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/HttpApplicationExtensions.cs
davidfowl: The answer is "What does ASP.NET do?"

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/HttpApplicationOptions.cs
```diff
@@ -0,0 +1,138 @@
...
+
+    /// <summary>
+    /// Gets or sets the number of <see cref="HttpApplication"/> retained for reuse. In order to support modules and appplications that may contain state,
+    /// a unique instance is required for each request. This type should be set to the average number of concurrent requests expected to be seen.
+    /// </summary>
+    public int PoolSize { get; set; } = 100;
+
+    public void RegisterModule<T>(string name)
+         where T : IHttpModule
+        => Modules.Add(name, typeof(T));
```
CZEMacLeod, line 48: To allow code similar to `System.Web.HttpApplication.RegisterModule`, could we do

```cs
    public void RegisterModule<T>(string? name = null)
         where T : IHttpModule
        => Modules.Add(name ?? MakeUniqueModuleName(typeof(T)), typeof(T));

    private static string MakeUniqueModuleName(Type moduleType) => 
        $"DynamicModule_{moduleType.AssemblyQualifiedName}_{Guid.NewGuid()}".ToString(CultureInfo.InvariantCulture);
```

Alternatively, could we just add these to the `IServiceProvider` then use `GetServices` when collecting the module factories in `HttpApplicationPooledObjectPolicy`?

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/HttpApplicationMiddleware.cs
twsouthwick: It also will new up and register all the modules. This can be potentially expensive, so I went with a pool. If someone doesn't want to have a pool, they could set the PoolSize to 0 or create a new ObjectPool<> that doesn't save anything

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/RequestHttpApplicationFeature.cs
twsouthwick: good call. I rework it so it track that state inside

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/HttpApplicationOptions.cs
twsouthwick: sure - I've added that

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/HttpApplicationPooledObjectPolicy.cs
```diff
@@ -0,0 +1,255 @@
...
+            var modules = new (string, IHttpModule)[moduleFactories.Count];
+
+            for (var i = 0; i < moduleFactories.Count; i++)
+            {
+                var module = (IHttpModule)moduleFactories[i].Item2(sp, null);
+                modules[i] = (moduleFactories[i].Key, module);
+            }
+
+            app.Initialize(new(modules), _state, eventInitializer);
+
```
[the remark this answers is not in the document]
twsouthwick: With an `HttpApplicationOptions.RegisterModule(Type)` overload the dynamic module scenario could be handled by hooking into that.

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/HttpApplicationMiddleware.cs
davidfowl: I’m asking if pooling is required to be correct.

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/HttpApplicationExtensions.cs
twsouthwick: and the answer is "no" for these, but "yes" to the events after logging. I've added some exception handling in to take care of these (and expanded the tests to cover throwing in each of the events)

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/HttpApplicationMiddleware.cs
twsouthwick: good call. I've reworked the exception handling to fit what is observed on framework (along with adding tests)

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/SessionMiddleware.cs
twsouthwick: same as above

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/HttpApplicationMiddleware.cs
twsouthwick: pooling is not required to be correct; there was no guarantee about an instance being reused.

twsouthwick:
@davidfowl would you prefer to start without pooling? Any other concerns to get this in?

CZEMacLeod:
@twsouthwick, @davidfowl 

To my mind, `System.Web` does this with pooling, and we now have excellent pooling mechanisms. As long as there are suitable tests for this, I don't see an issue.
I feel like the overheads, if you have a lot of modules/events and/or a complex Application class defined, are going to make this very slow without it.

If it does need to be disabled (at least by default, hopefully with an opt-in mechanism) could we do some kind of factory, and be able to register the `PoolingApplicationFactory` or a `SimpleApplicationFactory`. This would feel much better to me that just deleting the code outright.

It could allow other (customer supplied) implementations too by inheriting a base class, perhaps to spin up x instances at startup and anything that goes through `System.Web` is throttled to that many concurrent requests.

I don't want to make more work for you, and I want to see this next version get out (so we can get even more cool things added in), but if you are going to simplify this for the sake of maintainability/testability or otherwise, please make it a mechanism that can be overridden with a better implementation "at the consumer's risk".

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/HttpApplicationPooledObjectPolicy.cs
```diff
@@ -0,0 +1,255 @@
...
+    public HttpApplicationPooledObjectPolicy(IServiceProvider services, IOptions<HttpApplicationOptions> options, ILogger<HttpApplicationPooledObjectPolicy> logger)
+    {
+        _logger = logger;
+        _services = services;
+        _state = new HttpApplicationState();
+        _factory = new Lazy<Func<IServiceProvider, HttpApplication>>(() => CreateFactory(options.Value), isThreadSafe: true);
+    }
+
+    Action<IApplicationBuilder> IStartupFilter.Configure(Action<IApplicationBuilder> next) => builder =>
```
davidfowl, line 107: Why is the policy also the startup filter? Does it need to be coupled?

src/Microsoft.AspNetCore.SystemWebAdapters/HttpApplicationState.cs
```diff
@@ -0,0 +1,223 @@
...
+
+using System.Collections.Specialized;
+using System.Diagnostics.CodeAnalysis;
+using System.Threading;
+
+namespace System.Web;
+
+[SuppressMessage("Design", "CA1010:Generic interface should also be implemented", Justification = Constants.ApiFromAspNet)]
+[SuppressMessage("Design", "CA1001:Types that own disposable fields should be disposable", Justification = "This is internally disposed")]
+public sealed class HttpApplicationState : NameObjectCollectionBase
```
davidfowl, line 12: I'm assuming this is a straight port.

src/Microsoft.AspNetCore.SystemWebAdapters/HttpApplication.cs
```diff
@@ -0,0 +1,318 @@
...
+        {
+            RemoveEvent(eventName, handler);
+        }
+        else
+        {
+            throw new ArgumentOutOfRangeException(nameof(name));
+        }
+    }
+
+    private void AddEvent(ApplicationEvent appEvent, EventHandler handler)
```
davidfowl, line 272: This doesn't need to be thread safe?

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/HttpApplicationOptions.cs
```diff
@@ -0,0 +1,152 @@
...
+
+    public IDictionary<string, Type> Modules => _modules;
+
+    internal void MakeReadOnly() => _modules.MakeReadOnly();
+
+    /// <summary>
+    /// Gets or sets the number of <see cref="HttpApplication"/> retained for reuse. In order to support modules and appplications that may contain state,
+    /// a unique instance is required for each request. This type should be set to the average number of concurrent requests expected to be seen.
+    /// </summary>
+    public int PoolSize { get; set; } = 100;
```
davidfowl, line 47: What's the pool size on System.Web?

src/Microsoft.AspNetCore.SystemWebAdapters/HttpApplication.cs
twsouthwick: Nope - any instance of HttpApplication will be assigned to a request. If a new instance is required, then the initialization will occur on the request it is assigned to.

src/Microsoft.AspNetCore.SystemWebAdapters/HttpApplicationState.cs
twsouthwick: Yup - cleaned up a little to use more modern locking primitive, but overall it's the same

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/HttpApplicationPooledObjectPolicy.cs
twsouthwick: Nope - just implemented there instead of a new object

src/Microsoft.AspNetCore.SystemWebAdapters/HttpApplication.cs
twsouthwick: btw HttpApplication is per request, but the HttpApplicationState is shared across all requests

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/HttpApplicationOptions.cs
twsouthwick: There doesn't appear to be a specified number. They create on demand, then [this method](https://referencesource.microsoft.com/#System.Web/HttpApplicationFactory.cs,c535514db5112a4d,references) is called when there's idle time to trim 3% of the pool on each call

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/HttpApplicationPooledObjectPolicy.cs
twsouthwick: I've moved it to its own class

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/HttpApplicationOptions.cs
twsouthwick: This kind of dynamic pool size doesn't seem easy to implement on top of ObjectPool<> as-is, although I did play around with a custom implementation that would use ML.NET and its forecasting APIs to dynamically adapt the target pool size based on the historical concurrent request count :)

src/Microsoft.AspNetCore.SystemWebAdapters/HttpApplication.cs
davidfowl: I remember now 😬

davidfowl (approved):

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/HttpApplicationPooledObjectPolicy.cs
```diff
@@ -0,0 +1,234 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.ObjectPool;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.AspNetCore.SystemWebAdapters;
+
+internal sealed partial class HttpApplicationPooledObjectPolicy : PooledObjectPolicy<HttpApplication>, IDisposable
+{
+    private readonly HashSet<string> UnsupportedEvents = new(StringComparer.OrdinalIgnoreCase)
```
davidfowl, line 18: Why isn't this static?

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/HttpApplicationPooledObjectPolicy.cs
```diff
@@ -0,0 +1,234 @@
...
+    {
+        // Fired before the ASP.NET page framework sends content to a requesting client (browser).
+        "Application_PreSendContent",
+        
+        // Fired when the last instance of an HttpApplication class is destroyed. It's fired only once during an application's lifetime.
+        "Application_End",
+    };
+
+    private readonly Dictionary<string, Action<HttpApplication, EventHandler>> KnownEvents = new(StringComparer.OrdinalIgnoreCase)
```
davidfowl, line 27: Why isn;t this static.

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/HttpApplicationPooledObjectPolicy.cs
```diff
@@ -0,0 +1,234 @@
...
+        _state = new HttpApplicationState();
+        _factory = new Lazy<Func<IServiceProvider, HttpApplication>>(() => CreateFactory(options.Value), isThreadSafe: true);
+    }
+
+    public override HttpApplication Create()
+    {
+        var app = _factory.Value(_services);
+
+        // This is invoked each time an HttpApplication is constructed
+        app.InvokeEvent(ApplicationEvent.ApplicationInit);
```
davidfowl, line 109: I didn't realize this was an event. It would be useful to comment to the reference source or documentation about these things.

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/HttpApplicationPooledObjectPolicy.cs
```diff
@@ -0,0 +1,234 @@
...
+        return true;
+    }
+
+    private Func<IServiceProvider, HttpApplication> CreateFactory(HttpApplicationOptions options)
+    {
+        var eventInitializer = GetEventInitializer(options);
+        var factory = ActivatorUtilities.CreateFactory(options.ApplicationType, Array.Empty<Type>());
+        var moduleFactories = options.Modules
+            .Select(m => (m.Key, ActivatorUtilities.CreateFactory(m.Value, Array.Empty<Type>())))
+            .ToList();
```
davidfowl, line 126: Is this to support DI enabling HttpApplication?

twsouthwick: yes, as well as IHttpModule instances

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/HttpApplicationPooledObjectPolicy.cs
twsouthwick: Makes sense. I had a link at the feature level, but not here at the implementation level. Added a comment on the class so it's in the file and some links to equivalent stuff in reference source

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpApplication/HttpApplicationPooledObjectPolicy.cs
davidfowl: Is that a porting to the new system scenario though? I thought this was mostly for compat.

twsouthwick: The old system supported DI as well as of 4.7.2

davidfowl: Ah yes, that feature.....

PR #329, opened 2023-04-19
Add HttpApplication.VaryBy

twsouthwick:


twsouthwick:
@Tratcher This is to just add the API rather than #326 which adds helpers to tie it into the ASP.NET Core stack. Getting this in will help enable users to compile their modules while we figure out #326

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/SystemWebAdaptersExtensions.cs
```diff
@@ -122,6 +118,13 @@ public Action<IApplicationBuilder> Configure(Action<IApplicationBuilder> next)
             => builder =>
             {
                 builder.UseMiddleware<SetHttpContextTimestampMiddleware>();
+                builder.UseMiddleware<RegisterAdapterFeaturesMiddleware>();
+
+                if (builder.AreHttpApplicationEventsRequired())
+                {
+                    builder.UseMiddleware<HttpApplicationMiddleware>();
```
twsouthwick, line 125: This is being moved so that HttpApplication is available if someone wants to manually call the output caching before we get something in. Otherwise, the call to `app.UseOutputCaching()` would have to be after `app.UseSystemWebAdapters`

PR #343, opened 2023-05-26
Keep IPrincipal user instances if not ClaimsPrincipal

twsouthwick:
There are instances where people are using non-ClaimsPrincipal users in ASP.NET Core. This change will allow us to track that IPrincipal and surface it as needed as a ClaimsPrincipal.

This is mostly for use while migrating, as ideally they will switch to a ClaimsPrincipal based User, but will prevent information contained in the custom type from being lost.


src/Microsoft.AspNetCore.SystemWebAdapters/PrincipalUserFeature.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    public IPrincipal? User { get; set; }
+
+    ClaimsPrincipal? IHttpAuthenticationFeature.User
+    {
+        get => User switch
+        {
+            null => null,
+            ClaimsPrincipal claims => claims,
+            IPrincipal user => new ClaimsPrincipal(user),
```
Tratcher, line 21: Should this be cached?

Tratcher (approved):

src/Microsoft.AspNetCore.SystemWebAdapters/PrincipalUserFeature.cs
twsouthwick: For now, I don't think we need to. It's only used here in the case of a custom IPrincipal, which should be a somewhat fringe case and one we'd like them to move away from. I'll add some logging to flag these cases to le people know, as well as add a section in the docs to explain in more detail.

PR #345, opened 2023-05-27
Add build helpers for NuGet packages

twsouthwick:
This adds a target to the adapters that will allow making use of Content from older style NuGet packages that don't work with PackageReference. This is done by adding an attribute `CopyContent='true'` to the PackageReference node in the project file, which will cause all the content in the package to be passed into the Content ItemGroup and copied to the output directory.


src/Microsoft.AspNetCore.SystemWebAdapters/Build/Microsoft.AspNetCore.SystemWebAdapters.props
```diff
@@ -0,0 +1,3 @@
+<Project>
+  <PropertyGroup />
```
joperezr, line 2: I don't think you need an empty property group, just the open and closing tags of project are good enough

twsouthwick:
@jeffkl You helped me develop this a while back, and I'm looking to ship it in this library so that it's easy for customers to consume. Has anything changed on the NuGet side to make this not needed anymore or is it still valuable?

src/Microsoft.AspNetCore.SystemWebAdapters/Build/NuGetContentFiles.targets
```diff
@@ -0,0 +1,53 @@
...
+  <Target Name="AddNugetContentFiles" BeforeTargets="AssignTargetPaths">
+
+    <!-- Identity PackageReference and paths to packages -->
+    <ItemGroup>
+      <_PackageReferenceProperties Include="Pkg$([System.String]::Copy('%(PackageReference.Identity)').Replace('.','_'))">
+        <PackageName>%(PackageReference.Identity)</PackageName>
+        <CopyContent>%(PackageReference.CopyContent)</CopyContent>
+      </_PackageReferenceProperties>
+      <_PackageReferenceProperties Update="@(_PackageReferenceProperties)" PackageRootPath="$(%(_PackageReferenceProperties.Identity))" />
+      <_PackageReferenceCopyContent Include="@(_PackageReferenceProperties)" Condition="'%(CopyContent)' == 'true'" />
```
jeffkl, line 16: If users will always be using a new enough version of MSBuild you can use intrinsic item functions: https://learn.microsoft.com/en-us/visualstudio/msbuild/item-functions?view=vs-2022#intrinsic-item-functions
```suggestion
      <_PackageReferenceCopyContent Include="@(_PackageReferenceProperties->WithMetadataValue('CopyContent', 'true'))" />
```

Functions like `WithMetadataValue()` were added in recent version of 17.x though, so if a user has 15.x it would fail.  Just thought I'd mention it in case you can use it.

src/Microsoft.AspNetCore.SystemWebAdapters/Build/NuGetContentFiles.targets
```diff
@@ -0,0 +1,53 @@
...
+      <_ContentPkgProperties Include="@(_PackageReferenceCopyContent)" />
+
+      <!-- Filter out packages that need GeneratePathProperty -->
+      <_ValidContentPkgProperties Include="@(_ContentPkgProperties)" Condition=" '%(PackageRootPath)' != '' " />
+      <_InvalidContentPkgProperties Include="@(_ContentPkgProperties)" Condition=" '%(PackageRootPath)' == '' " />
+    </ItemGroup>
+
+    <Warning
+      Condition=" '@(_InvalidContentPkgProperties)' != '' "
+      Text="%(_InvalidContentPkgProperties.PackageName) does not contain a tools directory and thus needs 'GeneratePathProperty' == 'true'." />
```
jeffkl, line 38: Nit: I prefer this wording:

```suggestion
      Text="The PackageReference '%(_InvalidContentPkgProperties.PackageName)' must specify the attribute 'GeneratePathProperty="true"' when CopyContent is enabled." />
```

src/Microsoft.AspNetCore.SystemWebAdapters/Build/NuGetContentFiles.targets
```diff
@@ -0,0 +1,53 @@
...
+
+    <ItemGroup Condition=" '@(_ValidContentPkgProperties)' != '' ">
+      <_NuGetContentFiles Include="%(_ValidContentPkgProperties.PackageRootPath)\Content\**\*">
+        <BasePath>%(_ValidContentPkgProperties.PackageRootPath)\Content</BasePath>
+        <PackageIdentity>%(_ValidContentPkgProperties.PackageName)</PackageIdentity>
+      </_NuGetContentFiles>
+
+      <Content Include="@(_NuGetContentFiles)">
```
jeffkl, line 46: Some users hate that these content files show up in Solution Explorer and want this item to have `<Visible>false</Visible>`.

https://github.com/NuGet/Home/issues/4856

By default, NuGet shows them but I don't like that.  Just a heads up that people might ask for these to be hidden.

src/Microsoft.AspNetCore.SystemWebAdapters/Build/NuGetContentFiles.targets
twsouthwick: I don't know if we could make that assumption. Most likely, this only would be used for ASP.NET Core apps, but we currently support ASP.NET 6.

twsouthwick: I like that syntax better though - seems cleaner :)

src/Microsoft.AspNetCore.SystemWebAdapters/Build/NuGetContentFiles.targets
twsouthwick: I like that better too. Thanks

src/Microsoft.AspNetCore.SystemWebAdapters/Build/NuGetContentFiles.targets
twsouthwick: They currently don't show up solution explorer. I'm assuming that's because I have set the target `BeforeTargets="AssignTargetPaths"`.  Are there conditions in which it will show up I just haven't seen?

jeffkl: Oh sorry I thought I scrolled up and didn't see this in a target.  Only `Content` and `None` items outside of targets show up.

jeffkl (approved):

src/Microsoft.AspNetCore.SystemWebAdapters/Build/NuGetContentFiles.targets
```diff
@@ -0,0 +1,53 @@
+<Project>
+
+  <!--
+      Add NuGet content to the build output if specified:
+      <PackageReference Include="..." CopyContent="true" ... />
+       -->
+  <Target Name="AddNugetContentFiles" BeforeTargets="AssignTargetPaths">
+
+    <!-- Identity PackageReference and paths to packages -->
+    <ItemGroup>
```
jeffkl, line 10: + @rainersigwald so he can see this beautiful hackery 😄

Issue #347, opened 2023-05-30
Caching.Cache.Insert is missing the overload with 3 arguments


afshinm:
### Describe the bug

`Caching.Cache.Insert` is missing the overload with 3 arguments, see: https://learn.microsoft.com/en-us/dotnet/api/system.web.caching.cache.insert?view=netframework-4.8.1#system-web-caching-cache-insert(system-string-system-object-system-web-caching-cachedependency)

### To Reproduce

- Import System.Web
- Call `HttpRuntime.Cache.Insert(cacheKey, cacheValue, new System.Web.Caching.CacheDependency(path));`
- Compile to get the build error

### Further technical details 

Please include the following if applicable:

*ASP.NET Framework Application*:
- Technologies and versions used (i.e. MVC/WebForms/etc): MVC 5
- .NET Framework Version: 4.7.2
- IIS Version: 10.0.22621
- Windows Version: 10

PR #350, opened 2023-06-01
Adding Cache.Insert overload with 3 arguments

afshinm:
- [X] You've read the [Contributor Guide](../CONTRIBUTING.md) and [Code of Conduct](../CODE-OF-CONDUCT.md).
- [X] You've included unit tests for your change, where applicable.
- [ ] You've included inline docs for your change, where applicable.
- [X] There's an open issue for the PR that you are making. If you'd like to propose a new feature or change, please open an issue to discuss the change or find an existing issue.

**PR Title**

Adds missing Cache.Insert overload

**PR Description**
Caching.Cache.Insert is missing the overload with 3 arguments, see: https://learn.microsoft.com/en-us/dotnet/api/system.web.caching.cache.insert?view=netframework-4.8.1#system-web-caching-cache-insert(system-string-system-object-system-web-caching-cachedependency)

Addresses #347


test/Microsoft.AspNetCore.SystemWebAdapters.Tests/Caching/CacheTests.cs
```diff
@@ -103,6 +103,22 @@ public void InsertNoCallbacks()
         memCache.Verify(m => m.Set(key, item, It.Is<CacheItemPolicy>(e => e.AbsoluteExpiration.Equals(absoluteExpriration) && e.SlidingExpiration.Equals(slidingExpiration)), null), Times.Once);
     }
 
+    [Fact]
+    public void InsertNoAbsoluteSlidingExpiration()
+    {
+        // Arrange
+        var memCache = new Mock<MemoryCache>(_fixture.Create<string>(), null);
```
twsouthwick, line 110: What about non-null dependencies?

src/Microsoft.AspNetCore.SystemWebAdapters/Caching/Cache.cs
```diff
@@ -61,6 +61,17 @@ private static void AddChangeMonitors(CacheDependency? dependencies, CacheItemPo
 
     public void Insert(string key, object value) => _cache.Set(key, value, new CacheItemPolicy());
 
+    public void Insert(string key, object value, CacheDependency? dependencies)
+    {
+        var policy = new CacheItemPolicy
```
twsouthwick, line 66: What about the dependency argument? The original implementation is here: https://referencesource.microsoft.com/#System.Web/Cache/cache.cs,d083199dcfb5f0da - feel free to copy the implementation from there to here.

twsouthwick (changes requested):

src/Microsoft.AspNetCore.SystemWebAdapters/Caching/Cache.cs
afshinm: Is that why the build is failing btw?

twsouthwick: possibly

test/Microsoft.AspNetCore.SystemWebAdapters.Tests/Caching/CacheTests.cs
afshinm: CacheDependency cannot be null in this case, just fixed the method signature.

src/Microsoft.AspNetCore.SystemWebAdapters/Caching/CacheDependency.cs
```diff
@@ -68,13 +68,17 @@ internal CacheDependency()
 
     protected internal void FinishInit()
     {
-        HasChanged = changeMonitors.Any(cm => cm.HasChanged && (cm.GetLastModifiedUtc() > utcStart));
-        utcLastModified = changeMonitors.Max(cm => cm.GetLastModifiedUtc());
-        if (HasChanged)
+        if (changeMonitors.Count > 0)
```
afshinm, line 71: This check is because changeMonitors count can be 0 (See `Init()`)

src/Microsoft.AspNetCore.SystemWebAdapters/Caching/Cache.cs
afshinm: @twsouthwick `CacheDependency` argument is not currently being consumed in other methods but there is an unused private `AddChangeMonitors` that was introduced in this commit https://github.com/dotnet/systemweb-adapters/commit/66276e5b0cbd7c74d054684b34843f7b9c553127#diff-bfdc712f9e9845bfd447d4eb1cbbafdb48e1d076e57108d882eb3651699290a9R52 -- This is a good potential solution to consume and monitor the cache dependencies. 

I'm happy to refactor other methods as well but I wanted to double check with you first.

src/Microsoft.AspNetCore.SystemWebAdapters/Caching/CacheDependency.cs
twsouthwick: wouldn't this not reset HasChanged if it has potentially been set to something? Is this a correctness issue?

CZEMacLeod: changeMonitors.Any works and sets HasChanged if changeMonitors count is 0. However, utcLastModified might have an issue if the source list is empty. However, I think that the specific overload of Max for DateTime uses the generic version which uses default(T) in the case of the empty set rather than throwing an error. In this case, it should probably use the test, and set DateTime.MinValue otherwise. I think all the other paths just short circuit (e.g. the Foreach) when there are no items so I don't believe it causes any issue.

src/Microsoft.AspNetCore.SystemWebAdapters/Caching/Cache.cs
CZEMacLeod: @twsouthwick I know there was a little bit of code missing from the implementation that I found when I was doing the VirtualPathProvider implementation.
There are changes in my branch https://github.com/CZEMacLeod/systemweb-adapters/tree/czemacleod/vpp-1.1 that I think include this overload and also sort out some other stuff that was not fully implemented.
I've been meaning to complete the branch for VPP and submit a PR, but while I have a good test/sample implementation, I don't have any good explicit tests defined.
It would probably be good if I separated out all the Cache changes from the VPP changes, as they are fixes/enhancements rather than a new feature but I haven't had time to complete this work yet.
Let me see if I can update my branch and pull all the cache specific code out into a PR. 
One thing I found was that there were a few things (like HasChanged) that don't behave quite as logically as you would expect on the System.Web Cache implementation compared to the more generic version that was introduced later. I know that I did catch these inconsistencies and match them like for like in the update I did.

CZEMacLeod: I created #351 which contains all the changes I had queued for `Cache` and `CacheDependency`.
I've added the tests from this PR, and the missing overload for `Cache.Insert`. The test here for NoSliding/NoAbsolute is technically covered by [Cache.Insert(String, Object)](https://learn.microsoft.com/en-us/dotnet/api/system.web.caching.cache.insert?view=netframework-4.8.1#system-web-caching-cache-insert(system-string-system-object))
I also have tests when there is an actual dependent object (file or item).

src/Microsoft.AspNetCore.SystemWebAdapters/Caching/CacheDependency.cs
CZEMacLeod: I've included this fix in #351

twsouthwick:
@afshinm I've merged in #351 - if there's more that's needed for you, please rebase/merge and we can get it in.

afshinm:
@twsouthwick sounds good. thanks.

PR #353, opened 2023-06-08
Expose extension methods to wrap and unwrap adapters

twsouthwick:
As it currently stands, we only support using casting to convert between the two representations. This change exposes the extension methods we are using internally to expose the ASP.NET Core or System.Web objects.

Fixes https://github.com/dotnet/aspnetcore/issues/51119


twsouthwick:
@davidfowl Here are the extension methods. I'm not happy with the name, though, as it feels fairly generic (but maybe sufficient?).

src/Microsoft.AspNetCore.SystemWebAdapters/SystemWebAdapterExtensions.cs
```diff
@@ -2,47 +2,38 @@
...
-using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Web;
 using Microsoft.AspNetCore.Http.Features;
-using Microsoft.AspNetCore.SystemWebAdapters.Internal;
 
 namespace Microsoft.AspNetCore.SystemWebAdapters;
 
-internal static class InternalSystemWebExtensions
+public static class SystemWebAdapterExtensions
```
davidfowl, line 11: Do we make the method names more specific? Seeing GetAdapter looks a little hard to understand. 

-  `GetHttpRequestAdapter()`
- `GetSystemWebRequest()`

Feels API review worthy.

twsouthwick: reviewed and updated

src/Microsoft.AspNetCore.SystemWebAdapters/SystemWebAdaptersConversionExtensions.cs
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    public static HttpResponse AsSystemWeb(this HttpResponseCore response)
+    {
+        ArgumentNullException.ThrowIfNull(response);
+
+        return response.HttpContext.AsSystemWeb().Response;
+    }
+
+    public static HttpResponseCore AsAspNetCore(this HttpResponse response)
```
Tratcher, line 25: Sort these either by framework or by type (request/response/context)

src/Microsoft.AspNetCore.SystemWebAdapters/SystemWebAdaptersConversionExtensions.cs
```diff
@@ -0,0 +1,75 @@
...
+    {
+        ArgumentNullException.ThrowIfNull(response);
+
+        return response;
```
Tratcher, line 29: This relies on the implicit converter? But going the other way the implicit converters rely on the extensions? Be consistent.

src/Microsoft.AspNetCore.SystemWebAdapters/SystemWebAdaptersConversionExtensions.cs
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    public static HttpContextCore AsAspNetCore(this HttpContext context)
+    {
+        ArgumentNullException.ThrowIfNull(context);
+
+        return context;
+    }
+
+    public static HttpRequestCore GetAspNetCoreRequest(this HttpRequest request)
```
Tratcher, line 54:
```suggestion
    public static HttpRequestCore AsAspNetCore(this HttpRequest request)
```

Tratcher (changes requested):

src/Microsoft.AspNetCore.SystemWebAdapters/SystemWebAdaptersConversionExtensions.cs
twsouthwick: done

src/Microsoft.AspNetCore.SystemWebAdapters/SystemWebAdaptersConversionExtensions.cs
twsouthwick: thanks - must have missed that one

src/Microsoft.AspNetCore.SystemWebAdapters/SystemWebAdaptersConversionExtensions.cs
twsouthwick: It needs a private field to do the conversion, so I've made the implicit operators the source of truth.

twsouthwick: switched to an internal property

src/Microsoft.AspNetCore.SystemWebAdapters/HttpContextBase.cs
```diff
@@ -57,7 +57,7 @@ public virtual IPrincipal User
         public virtual object? GetService(Type serviceType) => throw new NotImplementedException();
 
         [return: NotNullIfNotNull(nameof(context))]
-        public static implicit operator HttpContextBase?(HttpContextCore? context) => context?.GetAdapterBase();
+        public static implicit operator HttpContextBase?(HttpContextCore? context) => context?.AsSystemWebBase();
```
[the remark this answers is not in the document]
twsouthwick: There's no way to get a HttpContext from HttpContextBase. If we were to do this, we'd have to rework everything on top of HttpContextBase

src/Microsoft.AspNetCore.SystemWebAdapters/HttpRequest.cs
```diff
@@ -290,7 +290,7 @@ private static void WriteTo(Stream source, Stream destination)
         public void Abort() => _request.HttpContext.Abort();
 
         [return: NotNullIfNotNull(nameof(request))]
-        public static implicit operator HttpRequest?(HttpRequestCore? request) => request.GetAdapter();
+        public static implicit operator HttpRequest?(HttpRequestCore? request) => request?.HttpContext.AsSystemWeb().Request;
 
         [return: NotNullIfNotNull(nameof(request))]
         public static implicit operator HttpRequestCore?(HttpRequest? request) => request?._request;
```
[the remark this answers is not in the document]
twsouthwick: I've swapped this per our discussion

src/Microsoft.AspNetCore.SystemWebAdapters/FeatureCollectionExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using Microsoft.AspNetCore.Http.Features;
+
+namespace Microsoft.AspNetCore.SystemWebAdapters;
+
+internal static class FeatureCollectionExtensions
+{
+    internal static TFeature GetRequired<TFeature>(this IFeatureCollection features)
```
Tratcher, line 11: Is GetRequiredFeature not available?
https://github.com/dotnet/aspnetcore/blob/64f95ef2377403e075cced875363419755d317a4/src/Extensions/Features/src/FeatureCollectionExtensions.cs#L21

Tratcher (approved):

src/Microsoft.AspNetCore.SystemWebAdapters/FeatureCollectionExtensions.cs
twsouthwick: We're targeting .NET 6 where it is not. I've looked into switching to that on .NET 7+, but we have internals visible to and it ends up causing compiler warnings of multiple resolved types.

PR #357, opened 2023-06-28
Forward IHtmlString on .NET 8

twsouthwick:
This adds a .NET 8 build target to the adapters so we can forward our implementation of IHtmlString to the new in-box version. With this, someone can use IHtmlString in .NET Standard, but will be forwarded on platforms that support it out of the box (i.e. .NET Framework and .NET 8+).

This change also reacts to some nullability changes that occured in .NET 8


twsouthwick:
@Tratcher Is there a reason we would want to wait for a .net 8 build in a released package?

Tratcher:
> @Tratcher Chris Ross FTE Is there a reason we would want to wait for a .net 8 build in a released package?

If this new API is available in previews that's good enough.

Tratcher (approved):

PR #358, opened 2023-06-28
Marking packages as stable in preparation for 1.2 release

joperezr:
FYI @twsouthwick @adityamandaleeka


adityamandaleeka (approved):
LGTM. Looking at the history, did we not do this for 1.1?

joperezr:
Good question, I hadn't noticed. Technically speaking you don't necesarilly need this change as you can queue a build and then explicitly pass the argument `/p:DotNetFinalVersionKind=release` through AzDOs options.

PR #359, opened 2023-06-28
Fix reliability of a couple of caching tests

twsouthwick:
These tests relied on some timing for file writes and static property access. This causes a race condition where the change monitor is disposed at some point internally, but at the same time the test tries to dispose it. This causes the list of change monitors to be enumerated/modified at the same time leading to a crash. The following are fixes/mitigations:

-  Used a TaskCompletionSource for better management of timing in the test to ensure we wait the right amount of time
- Added a lock in the dispose method to guard against reentrancy

This also changed the IHttpRuntime access to go through HttpContext.Current instead of a static property. While not necessary for the current bug, it has been the source of other issues where a test will set the static property at the same time another test is using it. Switching to HttpContext.Current will prevent similar issues as the IHttpRuntime will be request/test local.

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters/Caching/CacheDependency.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters/Caching/CacheDependency.cs
@@ -137,16 +137,22 @@ public class CacheDependency : IDisposable
         {
             if (disposing)
             {
+                // Ensure that if the Dispose is called by different threads (i.e. the wrapping ChangeMonitor) that it won't enumerate the list at the same time
                 lock (changeMonitors)
                 {
-                    foreach (var changeMonitor in changeMonitors)
+                    if (!disposedValue)
                     {
-                        changeMonitor?.Dispose();
-                    }
-                    changeMonitors.Clear();
-                }
+                        foreach (var changeMonitor in changeMonitors)
+                        {
+                            changeMonitor?.Dispose();
+                        }
+                        changeMonitors.Clear();
 
-                DependencyDispose();
+                        DependencyDispose();
+
+                        disposedValue = true;
+                    }
+                }
             }
 
             disposedValue = true;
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.SystemWebAdapters/Caching/CacheDependency.cs
```diff
@@ -135,14 +137,27 @@ protected virtual void Dispose(bool disposing)
         {
             if (disposing)
             {
-                foreach (var changeMonitor in changeMonitors)
+                if (!disposedValue)
```
joperezr, line 140: NIT: Do we actually need the three checks for `!disposedValue` or could we get away with just two by reordering the first check? Just mentioning it since typical locks usually only do two checks.

test/Microsoft.AspNetCore.SystemWebAdapters.Tests/Caching/CacheTests.cs
```diff
@@ -349,23 +353,25 @@ void Callback(string key, CacheItemUpdateReason reason, out object? expensiveObj
...
 
-        // Small delay here to ensure that the file change notification happens (may fail tests if too fast)
-        await Task.Delay(10);
+        // Wait for callback to be called
+        await tcs.Task;
 
-        // Force cleanup to initiate callbacks on current thread
-        memCache.Trim(100);
+        // Wait for callback to be finalized
+        await Task.Delay(100);
```
joperezr, line 374: Is there a risk here for flakiness on slower machines?

joperezr (approved):
Left a couple of very small nitpicks but looks great otherwise. Thanks for the fast turnaround @twsouthwick, hopefully we can get this in sooner rather than later so that we can push the packages to NuGet.

src/Microsoft.AspNetCore.SystemWebAdapters/Caching/CacheDependency.cs
twsouthwick: Yup - that's what I meant to do. I'll fix that up

test/Microsoft.AspNetCore.SystemWebAdapters.Tests/Caching/CacheTests.cs
twsouthwick: maybe - but it only needs to return from the callback method and a couple of operations so I think it should be fine.

PR #389, opened 2023-08-16
Use an options type for hosting and runtime environment

### LINKED ISSUE #368: Ask About System.Web.Hosting.HostingEnvironment
## Summary
There are some APIs of HostingEnvironment missing in .NetCore. I want to ask if there are equivalents and if there is plan to add them to adapter.

## Motivation and goals
We need migrate code (an API in Sharepoint) to ASP.NetCore and we find that the APIs are not supported by ASP.NetCore. These APIs seem not supported by IHostEnvironment.

## In scope

- System.Web.Hosting.HostingEnvironment.ApplicationID
- System.Web.Hosting.HostingEnvironment.IsHosted
- System.Web.Hosting.HostingEnvironment.SiteName
- System.Web.Hosting.HostingEnvironment.MapPath

## Examples

1. Use applicationId to find objects of the application
```C#
var appId = System.Web.Hosting.HostingEnvironment.ApplicationID;
var targetObject = Database.Objects.First(o =>o.Id == appId)
```

2. Get User from context when it is hosted
```C#
if (System.Web.Hosting.HostingEnvironment.IsHosted)
{
    HttpContext ctx= HttpContext.Current;
    if (ctx!= null)
    {
        return ctx.User;
    }
}
```

3. Use SiteName to find objects of a site
```C#
var siteName= System.Web.Hosting.HostingEnvironment.SiteName;
var targetObject = Database.Objects.Where(o =>o.SiteName == siteName)
```

4. If the physicalpath (filepath) does not exist, return
```C#
if (File.Exists(HostingEnvironment.MapPath("/targetPage.aspx")))
    return;
```

twsouthwick:
This refactors the runtime environment to be configured via a user-available options type. By default, if running on IIS, this will be populated, but can easily be configured if needed.

This also adds HostingEnvironment properties that may be expected via the same options.

As part of this, a hosted service is added that will set the options as a static instance so it will be available for the lifetime of the service (since the APIs on HttpRuntime and HostingEnvironment were static). This hosted service will take care of clearing this value once the host is stopped.

Fixes #368
Blocked by https://github.com/dotnet/aspnetcore/issues/43632 for full support


twsouthwick:
@Tratcher this enables the HostingEnvironment APIs on the adapter side. Once https://github.com/dotnet/aspnetcore/issues/43632 is available, we can hook it up as a separate PR. This enables customizing the values surfaced through the APIs, though, so can be useable without the full IIS hookup.

PR #391, opened 2023-08-24
Add IHttpRequestPathFeature.CurrentExecutionFilePath

### LINKED ISSUE #372: Ask About HttpRequest.CurrentExecutionFilePath
## Summary

Our repo uses HttpRequest.CurrentExecutionFilePath  which is not supported by ASP.NetCore and this adapters.
I want to ask if there is equivalent and if there is plan to add them to adapter.

## Motivation and goals

We need migrate code (an API in Sharepoint) to ASP.NetCore and we find that the APIs are not supported by ASP.NetCore.

## In scope

- HttpRequest.CurrentExecutionFilePath

From [document](https://learn.microsoft.com/en-us/dotnet/api/system.web.httprequest.currentexecutionfilepath?view=netframework-4.8.1), [CurrentExecutionFilePath](https://learn.microsoft.com/en-us/dotnet/api/system.web.httprequest.currentexecutionfilepath?view=netframework-4.8.1) returns the file path to the currently executing page handler. For redirect scenarios using [Execute](https://learn.microsoft.com/en-us/dotnet/api/system.web.httpserverutility.execute?view=netframework-4.8.1) and [Transfer](https://learn.microsoft.com/en-us/dotnet/api/system.web.httpserverutility.transfer?view=netframework-4.8.1) methods this means that the [CurrentExecutionFilePath](https://learn.microsoft.com/en-us/dotnet/api/system.web.httprequest.currentexecutionfilepath?view=netframework-4.8.1) property returns the path to the page redirected to (child page). However, when the client is redirected to another page, the [FilePath](https://learn.microsoft.com/en-us/dotnet/api/system.web.httprequest.filepath?view=netframework-4.8.1) property returns the path to the original page.

But in `HttpRequestInputStreamFeature.cs`, the `Rewrite` function shows that the FilePath is changed to the new path. There is no variable recording the original path.

I wonder that will adapters support it?

## Examples
```C#
bool isErrorPage = false;
if (context.Request.CurrentExecutionFilePath.EndsWith("Error.aspx"))
{
    isErrorPage = true;
}
```

Clounea:
Closes #372 

Since the Execute/Transfer is not supported yet. I don't add tests for CurrentExecutionFilePath.

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpRequestInputStreamFeature.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpRequestInputStreamFeature.cs
@@ -215,6 +215,8 @@ internal class HttpRequestInputStreamFeature : IHttpRequestInputStreamFeature, I
 
     string IHttpRequestPathFeature.RawUrl => _other.RawTarget;
 
+    string IHttpRequestPathFeature.CurrentExecutionFilePath => _filePath ?? Path;
+
     internal static class AspNetCoreTempDirectory
     {
         private static string? _tempDirectory;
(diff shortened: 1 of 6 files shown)


twsouthwick (approved):

twsouthwick:
Thanks @Clounea!

Issue #393, opened 2023-08-30
HttpCookie urlencodes cookie values


afshinm:
### Describe the bug

Create a new [`HttpCookie`](https://github.com/dotnet/systemweb-adapters/blob/2b29d699a18a64817e554452b255e52c3f3b59a3/src/Microsoft.AspNetCore.SystemWebAdapters/HttpCookie.cs#L10) object in NET Standard / NET Core URL encodes the value of of cookie, whereas in NET Framework 4.7.2, this doesn't occur. This is probably the desired behavior but makes it difficult to share the same cookie between the two apps (NET Core and NET Framework)

### To Reproduce

```
var cookie = new HttpCookie(cookieName, "a|b") { Secure = true, SameSite = SameSiteMode.None };
Console.WriteLine(cookie.Value); // prints a%7Cb
```

### Exceptions (if any)

None

### Further technical details 

A solution is to manually create the `Set-cookie` header in the request payload but that's very error prone and hard to test.

Please include the following if applicable:

*ASP.NET Framework Application*:
- Technologies and versions used (i.e. MVC/WebForms/etc): MVC
- .NET Framework Version: 4.7.2
- IIS Version: 10
- Windows Version: 11

*ASP.NET Core Application*:
- Targeted .NET version: .NET 6

joperezr:
[Triage]: We looked at this during triage and this seems to be a bug in the implementation. https://github.com/dotnet/systemweb-adapters/blob/main/src/Microsoft.AspNetCore.SystemWebAdapters/HttpCookie.cs  Seems like a good place to start for this investigation is checking the flags that may be causing the url encoding to happen [here](https://github.com/dotnet/systemweb-adapters/blob/2b29d699a18a64817e554452b255e52c3f3b59a3/src/Microsoft.AspNetCore.SystemWebAdapters/HttpCookie.cs#L82)

afshinm:
@joperezr Thanks for looking into this. Yes, more than happy to work on this. 

Just to double check my understanding, do you agree that the cookie values shouldn't be URL encoded automatically? At least that's the behavior I'm seeing in Net Framework 4.7.2.

afshinm:
[...]

PR #394, opened 2023-08-31
HttpCookie.Value should return the URL decoded cookie value

afshinm:
- [X] You've read the [Contributor Guide](../CONTRIBUTING.md) and [Code of Conduct](../CODE-OF-CONDUCT.md).
- [X] You've included unit tests for your change, where applicable.
- [X] You've included inline docs for your change, where applicable.
- [X] There's an open issue for the PR that you are making. If you'd like to propose a new feature or change, please open an issue to discuss the change or find an existing issue.

**PR Title**
HttpCookie.Value should return the URL decoded cookie value

**PR Description**

Create a new [`HttpCookie`](https://github.com/dotnet/systemweb-adapters/blob/2b29d699a18a64817e554452b255e52c3f3b59a3/src/Microsoft.AspNetCore.SystemWebAdapters/HttpCookie.cs#L10) object in NET Standard / NET Core URL encodes the value of of cookie, whereas in NET Framework 4.7.2, this doesn't occur. This is probably the desired behavior but makes it difficult to share the same cookie between the two apps (NET Core and NET Framework).

To reproduce this bug:

```
var cookie = new HttpCookie(cookieName, "a|b") { Secure = true, SameSite = SameSiteMode.None };
Console.WriteLine(cookie.Value); // prints a%7Cb
```

Also, here is an example that `HttpCookie` should just return the cookie value without URL encoding it: https://dotnetfiddle.net/eusU6N (but `Values` should URL encode, which I added a unit test for it in this PR).

Addresses #393

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters/HttpCookie.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters/HttpCookie.cs
@@ -9,9 +9,8 @@ namespace System.Web;
 
 public sealed class HttpCookie
 {
-    // If the .Values collection hasn't been accessed, this will remain a string. However, once .Values is accessed,
-    // this will become a HttpValueCollection. If .ToString is called on it, it will reconsistute the full string
-    private object? _holder;
+    private string? _stringValue;
+    private HttpValueCollection? _multiValue;
 
     public HttpCookie(string name)
     {
@@ -21,7 +20,7 @@ public sealed class HttpCookie
     public HttpCookie(string name, string? value)
     {
         Name = name;
-        _holder = value;
+        _stringValue = value;
     }
 
     /// <summary>
@@ -42,21 +41,30 @@ public sealed class HttpCookie
 
     /// <summary>
     /// Gets or sets an individual cookie value.
+    /// </summary>
     public string? Value
     {
-        get => _holder?.ToString();
+        get
+        {
+            if (_multiValue != null)
+                return _multiValue.ToString(false);
+            else
+                return _stringValue;
+        }
 
         set
         {
-            if (_holder is HttpValueCollection values)
+            if (_multiValue != null)
             {
-                // reset multivalue collection to contain single keyless value
-                values.Clear();
-                values.Add(null, value);
+                // reset multivalue collection to contain
+                // single keyless value
+                _multiValue.Clear();
+                _multiValue.Add(null, value);
             }
             else
             {
-                _holder = value;
+                // remember as string
+                _stringValue = value;
             }
         }
     }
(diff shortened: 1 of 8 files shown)


src/Microsoft.AspNetCore.SystemWebAdapters/Internal/HttpValueCollection.cs
```diff
@@ -89,7 +89,7 @@ internal void FillFromString(string s, bool urlencoded = false, Encoding? encodi
 
     public override string ToString() => ToString(true);
 
-    private string ToString(bool urlencoded)
+    internal string ToString(bool urlencoded)
```
afshinm, line 92: This is because `_multiValue.ToString(false);` needs to be called in `HttpCookie`. This is also `internal` in the reference source: https://github.com/microsoft/referencesource/blob/51cf7850defa8a17d815b4700b67116e3fa283c2/System.Web/HttpValueCollection.cs#L294C38-L294C38

afshinm:
I manually tested these changes locally and confirmed that `HttpCookie.Value` is now correctly reflecting the URL decoded value, however, the `Set-cookie` header generated in the response gets URL encoded again.

I'm not quite sure why this is happening or how the `Set-cookie` header gets generated but I noticed that this file generates the `Set-cookie` header https://github.com/microsoft/referencesource/blob/51cf7850defa8a17d815b4700b67116e3fa283c2/System.Web/HttpCookie.cs#L444. Maybe we should have the same method?

twsouthwick:
> I'm not quite sure why this is happening or how the `Set-cookie` header gets generated but I noticed that this file generates the `Set-cookie` header https://github.com/microsoft/referencesource/blob/51cf7850defa8a17d815b4700b67116e3fa283c2/System.Web/HttpCookie.cs#L444. Maybe we should have the same method?

So it sounds like your fix gets us partially there, but there is still some URL encoding going on? If so, would you be up for integrating that method in? There's some e2e tests that set up a test host that should allow testing this scenario

src/Microsoft.AspNetCore.SystemWebAdapters/HttpCookie.cs
```diff
@@ -68,37 +76,38 @@ public NameValueCollection Values
...
+            if (_multiValue == null)
             {
-                return values;
+                // create collection on demand
+                _multiValue = new HttpValueCollection();
+
+                // convert existing string value into multivalue
+                if (_stringValue != null)
+                {
+                    if (_stringValue.Contains('&', StringComparison.InvariantCulture) || _stringValue.Contains('=', StringComparison.InvariantCulture))
```
twsouthwick, line 87: is this what system.web was doing?

twsouthwick (approved):
LGTM - let me know if you want to merge this in or do the more extensive set-cookie fix

src/Microsoft.AspNetCore.SystemWebAdapters/HttpCookie.cs
afshinm: @twsouthwick no, System.web is doing an `.IndexOf` without the `SystemComparison` argument. Do you want me to change it to IndexOf?

afshinm:
> > I'm not quite sure why this is happening or how the `Set-cookie` header gets generated but I noticed that this file generates the `Set-cookie` header https://github.com/microsoft/referencesource/blob/51cf7850defa8a17d815b4700b67116e3fa283c2/System.Web/HttpCookie.cs#L444. Maybe we should have the same method?
> 
> So it sounds like your fix gets us partially there, but there is still some URL encoding going on? If so, would you be up for integrating that method in? There's some e2e tests that set up a test host that should allow testing this scenario

@twsouthwick Thanks for the review. 

I think I prefer to run some more e2e tests as this is not going to actually fix #393. Re e2e tests, are you referring to these Playwright tests here https://github.com/dotnet/systemweb-adapters/tree/main/test/Samples.MVCApp.Tests?

twsouthwick:
I was referring to these tests: https://github.com/dotnet/systemweb-adapters/blob/main/test/Microsoft.AspNetCore.SystemWebAdapters.CoreServices.Tests/ResponseStreamTests.cs

afshinm:
@twsouthwick Updated the PR and added the `ToSetCookieHeaderValue` method which correctly sets the `Set-cookie` headers. Also added some end to end tests. Let me know what you think.

src/Microsoft.AspNetCore.SystemWebAdapters/HttpCookie.cs
twsouthwick: No this is fine -I was wondering if the rest of it was something that system.web was doing

test/Microsoft.AspNetCore.SystemWebAdapters.CoreServices.Tests/ResponseHeaderTests.cs
```diff
@@ -0,0 +1,148 @@
...
+
+[Collection(nameof(SelfHostedTests))]
+public class ResponseHeaderTests
+{
+    private readonly string ContentValue = Guid.NewGuid().ToString();
+
+    [Fact]
+    public async Task SetCookie()
+    {
+        var result = await RunAsync(context =>
```
twsouthwick, line 28: nit: httpresponsemessage should have a `using`; not sure why there's no analyzer catching that

twsouthwick (commented):
Can you add the usings?

test/Microsoft.AspNetCore.SystemWebAdapters.CoreServices.Tests/ResponseHeaderTests.cs
afshinm: @twsouthwick good catch. I just updated the PR and added `using`. I can update the analyzer to catch this type of errors in a separate PR.

twsouthwick (approved):

PR #404, opened 2023-09-20
Adding more browser capabilities

birojnayak:
- [y ] You've read the [Contributor Guide](../CONTRIBUTING.md) and [Code of Conduct](../CODE-OF-CONDUCT.md).
- [ n/A] You've included unit tests for your change, where applicable.
- [ n/a] You've included inline docs for your change, where applicable.
- [ y] There's an open issue for the PR that you are making. If you'd like to propose a new feature or change, please open an issue to discuss the change or find an existing issue.

**PR Title**
Adding more browser capabilities

**PR Description**
Detail 1
Detail 2

Addresses #bugnumber (in this specific format)

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters/Configuration/HttpCapabilitiesBase.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters/Configuration/HttpCapabilitiesBase.cs
@@ -53,6 +53,14 @@ public class HttpCapabilitiesBase
 
     public string? this[string key] => Capability[key];
 
+    public Version EcmaScriptVersion => GetVersion("ecmascriptversion");
+
+    public Version MSDomVersion => GetVersion("msdomversion");
+
+    public Version W3CDomVersion => GetVersion("w3cdomversion");
+
+    public bool SupportsCallback => GetBoolean("supportsCallback");
+
     public bool IsMobileDevice => GetBoolean("isMobileDevice");
 
     private int GetInt(string key) => int.TryParse(Capability[key], NumberStyles.Integer, CultureInfo.InvariantCulture, out var result) ? result : throw new HttpUnhandledException($"Invalid string from browser capabilities '{key}'");
@@ -90,5 +98,18 @@ public class HttpCapabilitiesBase
         }
 
         throw new HttpUnhandledException($"Invalid string from browser capabilities '{key}'");
+    }
+
+    private Version GetVersion(string key)
+    {
+        var result = Capability[key];
+        if (result is not null)
+        {
+            return new Version(result);
+        }
+        else
+        {
+            throw new ArgumentNullException(nameof(key));
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.SystemWebAdapters/Configuration/HttpCapabilitiesBase.cs
```diff
@@ -53,6 +53,14 @@ private IHttpBrowserCapabilityFeature Capability
 
     public string? this[string key] => Capability[key];
 
+    public Version? EcmaScriptVersion => new Version(Capability["ecmascriptversion"]);
+
+    public Version? MSDomVersion => new Version(Capability["msdomversion"]);
+
+    public Version? W3CDomVersion => new Version(Capability["w3cdomversion"]);
+
+    public virtual bool SupportsCallback => GetBoolean("supportsCallback");
```
twsouthwick, line 62: let's remove the virtual here

src/Microsoft.AspNetCore.SystemWebAdapters/Configuration/HttpCapabilitiesBase.cs
```diff
@@ -53,6 +53,14 @@ private IHttpBrowserCapabilityFeature Capability
 
     public string? this[string key] => Capability[key];
 
+    public Version? EcmaScriptVersion => new Version(Capability["ecmascriptversion"] ?? string.Empty);
```
twsouthwick, line 56:
```suggestion
    public Version? EcmaScriptVersion =>Capability["ecmascriptversion"] ?? is {} v ? new Version(v) : null;
```

You'll probably want to make a helper method to do that. However, verify what .NET Framework did here - did it return a null (I'm assuming that based on your nullable annotation), or did it return a version of something?

twsouthwick (changes requested):

src/Microsoft.AspNetCore.SystemWebAdapters/Configuration/HttpCapabilitiesBase.cs
birojnayak: https://referencesource.microsoft.com/#mscorlib/system/version.cs,354  and here https://referencesource.microsoft.com/#mscorlib/system/version.cs,376 <= eventually if the version is null or empty or invalid it throws exception.  So I am making the API as not nullable to stick with framework and adding helper to handle as below. Let me know if looks below, I will submit the next revision.

```c#
    public Version EcmaScriptVersion => GetVersion("ecmascriptversion");
   
    private Version GetVersion(string key)
    {
        var result = Capability[key];
        if (result is not null)
        {
            return new Version(result);
        }
        else
        {
            throw new ArgumentNullException(nameof(key));
        }
    }
```

twsouthwick: If that's what framework is doing, that sounds good to me

twsouthwick (approved):

PR #409, opened 2023-10-04
Add HttpServerUtility.Transfer/Execute methods

### LINKED ISSUE #377: Ask About HttpServerUtility.TransferRequest
## Summary

Our repo uses `HttpServerUtility.TransferRequest` which is not supported by ASP.NetCore and this adapters.
I want to ask if there is equivalent and if there is plan to add them to adapter.

## Motivation and goals

We need migrate code (an API in Sharepoint) to ASP.NetCore and we find that the APIs are not supported by ASP.NetCore.

## In scope

`HttpServerUtility.TransferRequest`

From this [post](https://stackoverflow.com/questions/39693050/alternative-to-server-transfer-in-asp-net-core), it seems that there are several options.
Will adapters add this function?

## Examples
```C#
context.Server.TransferRequest(
    modifiedUri,
    true,
    null,
    context.Request.Headers);
```

twsouthwick:
This should help with some migrations that use the type to highlight they need to refactor. For now, there's no implementation and it will throw if used, as well as show warnings when compiling. However, if there's a way to implement it (and if there's enough demand) an internal interface is available that that can be implemented to light up the feature.

Fixes #377

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters/Constants.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters/Constants.cs
@@ -12,4 +12,22 @@ internal static class Constants
     internal const string DisposeIsRegistered = "Object is registered for dispose with HttpContext";
 
     internal const string CA1859 = "False positive fixed by https://github.com/dotnet/roslyn-analyzers/pull/6421 but not integrated in yet";
+
+    internal static class TransferRequest
+    {
+        public const string Message = "TransferRequest is not supported on ASP.NET Core";
+        public const string DiagnosticId = "SYSWEB0001";
+    }
+
+    internal static class Execute
+    {
+        public const string Message = "Execute is not supported on ASP.NET Core";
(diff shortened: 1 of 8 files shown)


src/Microsoft.AspNetCore.SystemWebAdapters/Constants.cs
```diff
@@ -12,4 +12,22 @@ internal static class Constants
     internal const string DisposeIsRegistered = "Object is registered for dispose with HttpContext";
 
     internal const string CA1859 = "False positive fixed by https://github.com/dotnet/roslyn-analyzers/pull/6421 but not integrated in yet";
+
+    internal static class TransferRequest
+    {
+        public const string Message = "TransferRequest is not supported on ASP.NET Core";
+        public const string DiagnosticId = "SYSWEB0001";
```
twsouthwick, line 19: Is there a different prefix we should use?

Tratcher (approved):

PR #411, opened 2023-10-04
Make features public to enable testing/customization scenarios

### LINKED ISSUE #332: How to set up mocks/stubs for testing?
Hello,

Let's say I have some old method and xunit test for it:
```
int MyMethod() {
   HttpContext.Current.Response.Cookies.Add("foo", "bar");
   HttpContext.Current.Response.Write("whatever");
   return 123;
}

[Fact]
void MyTest() {
   // stub
   HttpContext.Current = new HttpContext(new HttpRequest("page.aspx", "http://localhost/", null), new HttpResponse(new StringWriter()));

   // test code
   var result = MyMethod();
   Assert.Equal(123, result);
}
```

What is needed to make the test work in netcore with adapter?
I can rewrite the test, but not the method under test (that's the reason for using the adapter).

The only way I've found so far is to run a full-blown inmemory test server, similar to [this test in the repo](https://github.com/dotnet/systemweb-adapters/blob/main/test/Microsoft.AspNetCore.SystemWebAdapters.CoreServices.Tests/HttpContextIntegrationTests.cs).
But that's quite an overkill: I don't want to test ASP.NET features, I want to stub them.

twsouthwick:
This exposes the internal features we use to enable System.Web scenarios for two main reasons:

1. Some functionality (such as ITransferRequestFeature) is not implemented out of the box
2. Requests have been made to have access to this for mocking purposes

For now, these are marked as experimental using the new ExperimentalAttribute to give us flexibility to change them if we need to since we haven't had to use them outside of our codebase. They have been moved into the `Microsoft.AspNetCore.SystemWebAdapters.Features` namespace as part of this change.

Fixes #332


twsouthwick:
@joperezr I'm making use of the `ExperimentalAttribute` here - is there support for this for downlevel, or should we just enable it for .NET 8+?

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/Features/IHttpApplicationFeature.cs
```diff
@@ -4,13 +4,23 @@
...
-internal interface IHttpApplicationFeature
+/// <summary>
+/// Represents the emulated IIS pipeline and the <see cref="HttpApplication"/> associated with it.
+/// </summary>
+#if NET8_0_OR_GREATER
+[Experimental(Constants.ExperimentalFeatures.DiagnosticId)]
+public
+#else
+internal
+#endif
```
Tratcher, line 22: This is awkward. We should make these public on all versions and worry about Experimental separately. If there's no convenient Experimental support downlevel then use Obsolete("For testing only, may change in the future") instead.

joperezr:
> @joperezr I'm making use of the `ExperimentalAttribute` here - is there support for this for downlevel, or should we just enable it for .NET 8+?

The attribute itself is .NET 8 only, but the feature is built-in to the compiler so the suggestion for downlevel is just to define an internal attribute that looks just like the .NEt 8 one, and the compiler won't care in which assembly it is defined

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/HttpResponseAdapterFeature.cs
```diff
@@ -10,6 +10,7 @@
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
+using Microsoft.AspNetCore.SystemWebAdapters.Features;
```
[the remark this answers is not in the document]
twsouthwick: moved

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/Features/IHttpApplicationFeature.cs
twsouthwick: per @joperezr, I've added an internal experimental attribute

twsouthwick:
> > @joperezr I'm making use of the `ExperimentalAttribute` here - is there support for this for downlevel, or should we just enable it for .NET 8+?
> 
> The attribute itself is .NET 8 only, but the feature is built-in to the compiler so the suggestion for downlevel is just to define an internal attribute that looks just like the .NEt 8 one, and the compiler won't care in which assembly it is defined

Cool. I added an internal one for downlevel then. Since people may be using it on .NET 6 as well, will they not see any compiler warnings? Does it make sense to add an obsolete to that as well (with the diagnostic id?)?

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/Features/IHttpApplicationFeature.cs
twsouthwick: had a merge issue - they've been removed

src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/Features/IHttpRequestInputStreamFeature.cs
```diff
@@ -3,14 +3,19 @@
 
 #if NETCOREAPP
 
+namespace Microsoft.AspNetCore.SystemWebAdapters.Features;
+
+using System.Diagnostics.CodeAnalysis;
```
Tratcher, line 8: using before namespace

twsouthwick: I'd noticed an inconsistency, but with the file scoped namespaces, I wasn't sure what the pattern was. I updated .editorconfig to enforce that

Tratcher (approved):

PR #418, opened 2023-10-10
Add System.Web.Security.MachineKey

### LINKED ISSUE #324: Consider supporting `System.Web.Security.MachineKey`
## Summary

`System.Web.Security.MachineKey` is a static class that is used in class libraries that can be adapted to be used with ASP.NET Core applications.

## Motivation and goals

`MachineKey` calls are static calls that may occur in low levels of code that is used to protect/unprotect data. In ASP.NET Core, the data protection APIs provide similar functionality. However, as with HttpContext.Current, these calls have been static on ASP.NET Framework and now require DI to retrieve on ASP.NET Core.

My providing the APIs for `MachineKey`, code that relies on this can be migrated to .NET Standard, but get the benefits of data protection without having to refactor the existing code to pass it in. Given that [there is support in ASP.NET Framework](https://learn.microsoft.com/aspnet/core/security/data-protection/compatibility/replacing-machinekey) to replace the underlying provider for MachineKey with the data protection APIs, this feature will simplify migration for apps.

## In scope

- In `Microsoft.AspNetCore.SystemWebAdapters`

```cs
namespace System.Web.Security.MachineKey
{
  public static class MachineKey
  {
    public static byte[] Unprotect(byte[] protectedData, params string[] purpose);
    public static byte[] Protect(byte[] userData, params string[] purposes);
  }
}
```

## Out of scope

- We do not want to try to mimic the actual data protection that occurs in .NET Framework in-box. If someone wants to do that, it will integrate into the adapters as-is as long as it is the registered `IDataProtectionProvider`. To support this migration scenario, we want people to initially move their use of the protection to the new stack.

## Risks / unknowns

- Users may not understand that they must do some extra configuration in order to share keys between framework and core apps to be able to unprotect data the other protects
- In order to replace the data protection on framework, it will invalidate existing values that are protected and steps must be take to mitigate that

# Detailed design

An initial implementation and sample:

- [MachineKey implementatoin](https://github.com/dotnet/systemweb-adapters/blob/tasou/machine-key/src/Microsoft.AspNetCore.SystemWebAdapters/Security/MachineKey.cs)
- [Sample](https://github.com/dotnet/systemweb-adapters/tree/tasou/machine-key/samples/MachineKey) (video below)

https://user-images.githubusercontent.com/583206/232643859-9d7de94b-a6f9-488c-81b6-8767ea6b273f.mp4

twsouthwick:
This adds the API backed by the ASP.NET Core data protection APIs. In order to share code between ASP.NET Core and framework with the same behavior, the ASP.NET Core data protection APIs must be used on the framework side as well. A sample has been added that shows how to enable that (this is already a supported scenario for ASP.NET Framework - this PR just hooks it up so the System.Web APIs can be used if needed).

Fixes #324


samples/MachineKey/MachineKeyFramework/packages.config
```diff
@@ -0,0 +1,30 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="Microsoft.AspNetCore.Cryptography.Internal" version="2.2.0" targetFramework="net472" />
```
Tratcher, line 3: Use 2.1, not 2.2 for all of these.

samples/MachineKey/MachineKeyCore/Program.cs
```diff
@@ -0,0 +1,23 @@
...
+
+builder.Services.AddSystemWebAdapters();
+
+var app = builder.Build();
+
+app.UseSystemWebAdapters();
+
+app.Map("/", (HttpContext context) =>
+{
+    context.Features.GetRequiredFeature<IHttpBodyControlFeature>().AllowSynchronousIO = true;
```
Tratcher, line 19: Eeewww 😂

Tratcher (approved):
@blowdart FYI

samples/MachineKey/MachineKeyFramework/packages.config
twsouthwick: I just installed the package and it added these. Should we be delisting the 2.2.0 packages from nuget so people use the 2.1.0 versions?

twsouthwick: wow I don't miss packages.config :( I've updated them to 2.1

Tratcher: 2.2 is already flagged as deprecated, nuget _shouldn't_ recommend them.

PR #424, opened 2023-10-16
Use AddWrappedAspNetCoreSession to better fit standard naming

twsouthwick:
We have a method (WrapAsnnetCoreSession) that wraps the ASP.NET Core session for use with System.Web.SessionState APIs, but this name is not in line with the standard service collection extensions of (`Add...`). This change obsoletes the old one and applies the normal convention.

### DIFF
--- a/samples/CoreApp/Program.cs
+++ b/samples/CoreApp/Program.cs
@@ -1,7 +1,7 @@
 var builder = WebApplication.CreateBuilder();
 
 builder.Services.AddSystemWebAdapters()
-    .WrapAspNetCoreSession()
+    .AddWrappedAspNetCoreSession()
     .AddSessionSerializer()
     .AddCustomSerialization();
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/SessionState/Wrapped/WrappedAspNetCoreSessionExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+using System;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.SystemWebAdapters;
+using Microsoft.AspNetCore.SystemWebAdapters.SessionState.Serialization;
+using Microsoft.AspNetCore.SystemWebAdapters.SessionState.Wrapped;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+
+namespace Microsoft.Extensions.DependencyInjection;
+
+public static class WrappedAspNetCoreSessionExtensions
```
Tratcher, line 13: Why did it need a new class? The old one was fine.

Tratcher (approved):

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/SessionState/Wrapped/WrappedAspNetCoreSessionExtensions.cs
twsouthwick: reverted that change

Issue #449, opened 2023-11-25
Don't set content-type if not content


jherven:
## Summary

Using this library in a dotnet core application forcefully adds content-type "text/html" (when not set) even if there's no content. This isn't always desired and I think it's neither expected when there is no content for the response.

## Motivation and goals

- Some clients break when there is a content-type set but no content
- At least for my migrated projects this doesn't mimic the behavior I had in the old application

## Risks / unknowns

If somebody already using this library relies on this header always being set this change may break something

# Detailed design

The change I've done for now is very simple.

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/SetDefaultResponseHeadersMiddleware.cs -> WriteDefaultContentType

Added new criteria "hasContent"
```
bool hasContent = context.Response.ContentLength.HasValue;
        if (hasContent && context.Response.Headers.ContentType.Count == 0)
        {
            context.Response.Headers.ContentType = "text/html";
        }
```

twsouthwick:
Thanks @jherven - have you verified this is the same behavior as on ASP.NET Framework? If so, would you be willing to submit a PR?

jherven:
Yes, for the scenarios I tested this change better mimics ASP.NET framework. PR created #450 .

PR #450, opened 2023-11-27
Don't force default content-type if no content

jherven:
**Don't force default content-type if no content**
If response has no content, don't automatically add content-type header

**PR Description**
- If response has content nothing changes
- If response has no content don't automatically add any content-type header

Addresses #449


src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/SetDefaultResponseHeadersMiddleware.cs
```diff
@@ -32,7 +32,8 @@ public Task InvokeAsync(HttpContext context)
 
     private static void WriteDefaultContentType(HttpContext context)
     {
-        if (context.Response.Headers.ContentType.Count == 0)
+        bool hasContent = context.Response.ContentLength.HasValue;
+        if (hasContent && context.Response.Headers.ContentType.Count == 0)
```
twsouthwick, line 36:
```suggestion
        if (context.Response.ContentLength.HasValue && context.Response.Headers.ContentType.Count == 0)
```

twsouthwick (commented):
There are some tests that are probably failing here

twsouthwick:
Thanks @jherven for the fix!

PR #399, opened 2023-09-05
add support for context.SetSessionStateBehavior

Clounea:
close #371

I don't know if I could change the metadata. Comments are welcome.

And I don't find a good place to add tests. It seems that I need a web app and a beginrequesthandler where I could setsessionstatebehavior.

PR #463, opened 2024-01-09
Ensure SessionAttribute defaults to same behavior as 1.2

### LINKED ISSUE #455: System.Web.HttpContext.Session is null with SystemWebAdapters v1.3.0 on .Net 8 after upgrading the 1.20 version on .Net 6.
Originally I'd already created a project using .Net 6 and SystemWebAdapters using version 1.2.0 and it worked fine.

After migrating to the .Net 8 and SystemWebAdapters 1.3.0 (nuget package) Session from the System.Web.HttpContext is returned as a null reference.

Initialization code is:

services.AddSystemWebAdapters()
    .AddSessionSerializer()
    .AddMssAdminSessionSerializer()
    .AddRemoteAppClient(options =>
    {
        options.RemoteAppUrl = new(Configuration.GetMigrationLegacyAppAddress());
        options.ApiKey = Configuration.GetMigrationRemoteAPIKey();
    })
    .AddSessionClient();

app.MapRazorPages()
    .RequireSystemWebAdapterSession();

One thing I see that this github project refers 1.2.0 version only. It is possible the 1.3.0 has changed initialization steps or maybe just moved the source to another place?

So for now, I just downgraded the SystemWebAdapters packages back to 1.2.0 version. Confirmed the Session sharing works fine under .Net 8 in this case.

twsouthwick:
As part of expanding the session support to enable per request changing the session behavior in #399, it appears that the default `SessionAttribute` was changed to have a different default value.

Fixes #455

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/SessionAttribute.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/SessionAttribute.cs
@@ -33,7 +33,7 @@ public sealed class SessionAttribute : Attribute
         }
     }
 
-    public SessionStateBehavior SessionBehavior { get; set; }
+    public SessionStateBehavior SessionBehavior { get; set; } = SessionStateBehavior.Required;
 
     public bool IsPreLoad { get; set; } = true;
 
--- a/test/Microsoft.AspNetCore.SystemWebAdapters.CoreServices.Tests/SessionState/SessionIntegrationTests.cs
+++ b/test/Microsoft.AspNetCore.SystemWebAdapters.CoreServices.Tests/SessionState/SessionIntegrationTests.cs
@@ -31,6 +31,7 @@ public class SessionIntegrationTests
 
     [InlineData("/disabled", "Session:null")]
     [InlineData("/readonly", "ReadOnly:True")]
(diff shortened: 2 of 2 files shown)


twsouthwick:
@joperezr not sure who I should be adding for reviews now, but once this is merged, we should push a v1.3.1 to fix this and a couple other regressions that have been fixed

joperezr:
We haven't really defined who will be the dev working on this repo yet, but I'm happy to take a look at this one. Seems like it was a regression (reported by the community) and fix seems trivial and tested, so I'm supportive of this. Happy to ship a new version with this, but do note that main is currently 1.4.0 as opposed to 1.3.1. Do we care about it being 1.3.1?

joperezr (approved):

twsouthwick:
Following semantic versioning would put us at 1.3.1, but 1.4.0 is fine if it will cause problems to change the version

joperezr:
It shouldn't it is just a matter of one more update in Versions.props. Or alternatively creating a release/3.1 branch from the commit that released 1.3.0 and cherry-picking this PR there. Either option would be fine.

twsouthwick:
It would be great to just take all 4 commits since 1.3 and release it as 1.3.1 - however that needs to be done I'll leave up to you :)

PR #464, opened 2024-01-09
Use hosted environment for CacheDependency tests

twsouthwick:
This moves tests that require CacheDependency into CoreServices. The implementation of this requires HttpRuntime.Cache to be set in some code paths and so we need to use the SelfHostedTests collection to ensure this is setup correctly and doesn't stomp on other tests. The change looks bigger than it is because of code movement - the only real change is the change to DependentItemCallback to use an actual host (test was originally written before this pattern was used in the codebase).


twsouthwick:
@joperezr looks like the build servers might have changed as I saw a test failure on my recent change. It was a test issue so not a regression or anything, but this fixes that.

joperezr (approved):

PR #487, opened 2024-02-28
Add test to verify method signature of adapted types

twsouthwick:
Turns out there are a few types we didn't port correctly. This adds a test to catch further regressions, as well as to track a baseline of known issues.


test/Microsoft.AspNetCore.SystemWebAdapters.Apis.Tests/Microsoft.AspNetCore.SystemWebAdapters.Apis.Tests.csproj
```diff
@@ -0,0 +1,40 @@
...
+    <Using Include="System.Web.HttpRequest" Alias="HttpRequest" />
+    <Using Include="System.Web.HttpResponse" Alias="HttpResponse" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\Microsoft.AspNetCore.SystemWebAdapters\Microsoft.AspNetCore.SystemWebAdapters.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Update="BaselineOk.txt">
```
joperezr, line 30: What generates/checks these?

joperezr (approved):

test/Microsoft.AspNetCore.SystemWebAdapters.Apis.Tests/Microsoft.AspNetCore.SystemWebAdapters.Apis.Tests.csproj
twsouthwick: It's a manual process - i.e. the test identifies a bunch of APIs whose docids don't match. Generally, these are suspect that they won't resolve on framework. These can then be verified that they actually do work, so we can add them to the BaselineOk.txt file

PR #491, opened 2024-02-28
Add HostingEnvironment.ApplicationPhysicalPath/ApplicationVirtualPath

twsouthwick:


src/Microsoft.AspNetCore.SystemWebAdapters/SystemWebAdaptersOptions.cs
```diff
@@ -4,19 +4,46 @@
...
 using System;
+using System.Web;
+using System.Web.Hosting;
 
 namespace Microsoft.AspNetCore.SystemWebAdapters;
 
 public class SystemWebAdaptersOptions
 {
+    /// <summary>
+    /// Gets or sets the value used by <see cref="HostingEnvironment.ApplicationID"/>
```
joperezr, line 15: Interesting. I see that most of the code doesn't have tripple slash comments, which makes sense as we usually add things here for people that are already using the API, but for this one it makes sense as this is specific about registering the adapters. I wonder if there is a way to enforce docs on public API on the latter bucket but not on the former.

joperezr (approved):

src/Microsoft.AspNetCore.SystemWebAdapters/SystemWebAdaptersOptions.cs
twsouthwick: I'm not aware of anything in that area, but it would be nice

PR #498, opened 2024-03-28
Ensure *Base types have the same members as the main types

twsouthwick:
System.Web had *Base types that were meant for abstraction. This adds a test to ensure we have the same members on both versions and adds a few that were missing.


Microsoft.AspNetCore.SystemWebAdapters.slnf
```diff
@@ -6,12 +6,13 @@
       "src\\Microsoft.AspNetCore.SystemWebAdapters.CoreServices\\Microsoft.AspNetCore.SystemWebAdapters.CoreServices.csproj",
       "src\\Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices\\Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices.csproj",
       "src\\Microsoft.AspNetCore.SystemWebAdapters\\Microsoft.AspNetCore.SystemWebAdapters.csproj",
+      "test\\Microsoft.AspNetCore.SystemWebAdapters.Apis.Tests\\Microsoft.AspNetCore.SystemWebAdapters.Apis.Tests.csproj",
```
twsouthwick, line 9: I added this in a recent pr but forgot to add to this filter which is what is run on ci

twsouthwick:
@joperezr here's one similar to #487 that is to ensure future additions are consistent and has some tests to validate structure

joperezr (approved):

PR #510, opened 2024-07-11
Remove or fix APIs that were ported with different signatures

### LINKED ISSUE #486: Some methods don't resolve on .NET Framework
During porting, we had been using various tools (such as EnablePackageValidation and some tests) to validate that the members/types ported matched with System.Web.dll so that the assembly would resolve correctly. However, looks like the testing did not validate return types correctly. This is to track methods that will not resolve correctly on framework when using the .NET Standard library:

Can obsolete and add the correct members if possible:

- [ ] the following `HttpException` constructors should not exist:
  -T:System.Void M:System.Web.HttpException.#ctor(System.Int32)
  -T:System.Void M:System.Web.HttpException.#ctor(System.Net.HttpStatusCode)
  - T:System.Void M:System.Web.HttpException.#ctor(System.Net.HttpStatusCode,System.String)
  - T:System.Void M:System.Web.HttpException.#ctor(System.Net.HttpStatusCode,System.String,System.Exception)
- [ ] `HttpException.StatusCode` should be `int HttpException.GetHttpCode()`
- [ ] `CacheDependency.Dispose(bool)` should not exist
- [ ] `System.Web.Caching.Cache.GetEnumerator()` should be an implicit interface implementation

Will require a breaking change to the API:

- [ ] `HttpRequest.LogonUserIdentity` should be `WindowsIdentity`
- [ ] `HttpContextBase.Application` should use the type HttpApplicationStateBase
- [x] `HttpResponseBase.Cache` should be of type `System.Web.HttpCachePolicy` (#476)

Thanks @AvremelM for point out the HttpResponseBase issue. I've added a test in #487 to catch further regressions and to identify other existing ones.

This will probably necessitate a bump for v2.0 of the adapters as changing these signatures is technically a breaking change (although it is currently in a broken state...)

See [the baseline of known issues](https://github.com/dotnet/systemweb-adapters/blob/c7c9d4431317a290238383e272b489dc51050f6b/test/Microsoft.AspNetCore.SystemWebAdapters.Apis.Tests/BaselineMethodIssues.txt) that should be updated as these are fixed

twsouthwick:
This is to fix a few APIs where the APIs ported did not match .NET Framework, so using those methods would throw at runtime. The only API removed in this is HttpRequest.LogonUserIdentity - we can investigate adding it back if needed. The rest of the APIs now match framework and tests are in place to ensure going forward no regressions or repeats of this issue occur.

Fixes #486


src/Microsoft.AspNetCore.SystemWebAdapters/HttpCachePolicyBase.cs
```diff
@@ -12,17 +12,14 @@ public class HttpCachePolicyBase
 
     public virtual void SetCacheability(HttpCacheability cacheability) => throw new NotImplementedException();
 
-    public virtual void SetCacheability(HttpCacheability cacheability, string field) => throw new NotImplementedException();
-    
-    [SuppressMessage("Microsoft.Naming", "CA1716:IdentifiersShouldNotMatchKeywords", MessageId = "Date",
-        Justification = "Matches HttpCachePolicy class")]
+    [SuppressMessage("Naming", "CA1716:Identifiers should not match keywords", Justification = Constants.ApiFromAspNet)]
```
CZEMacLeod, line 15: Looks like you've removed the SetCacheability overload here... Was that intentional?

twsouthwick: Yes - it wasn't implemented on HttpCachePolicy itself so we couldn't forward it in the wrapper correctly. Happy to take a fix that implements it on all three as a follow up :) I've added tests to validate that the three (i.e. HttpContext/HttpContextBase/HttpContextWrapper) are all kept in sync when adding APIs and this case was causing failure

src/Microsoft.AspNetCore.SystemWebAdapters/HttpException.cs
```diff
@@ -17,50 +18,32 @@ public HttpException(string message) : base(message)
     {
     }
 
-    public HttpException(string message, Exception innerException)
+    public HttpException(String message, Exception innerException)
```
mjrousos, line 21: Super nit-pick but this change seems unnecessary. We use `string` everywhere else in this file.

mjrousos (approved):

PR #511, opened 2024-07-11
Update caching to v8.0.0 on .NET 8.0 build

### LINKED ISSUE #488: Upgrade System.Runtime.Caching to v8.0.0
We're currently still using v6.0.0. System.Runtime.Caching before v8.0.0 implicitly brought in System.Drawing.Common. This makes it so by using the adapters and not explicitly opting into v8.0.0+ developers will find they have access to System.Drawing.Common types. This is fine for windows deployments, but if someone wants to move to linux, they will not have the compile time checks to make sure they're not using unavailable types.

twsouthwick:
Fixes #488


mjrousos (approved):
Left a suggestion inline, but not blocking.

PR #512, opened 2024-07-11
Use built-in GetRequiredFeature when available

twsouthwick:
This change updates to use the in-box GetRequiredFeature. As part of this, the .NET 7 target is dropped for v2.0 since it's out of support and made building this tricky.


mjrousos (approved):
LGTM

PR #515, opened 2024-07-12
Add support for HttpApplication.PreSendRequestContent

twsouthwick:
In order to support this event, we need to hook into the point at which content is flushed to the client. To achieve this, this change adds the following:

- A new implementation of IHttpResponseBodyFeature that gets added early when modules are registered that will intercept calls to FlushAsync
- Moves the PreSendRequestHeaders to the same interception point
- Reworks tests to run all the module tests for both buffered and non-buffered streams

### DIFF
--- a/samples/Modules/ModulesLibrary/BaseModule.cs
+++ b/samples/Modules/ModulesLibrary/BaseModule.cs
@@ -12,7 +12,7 @@ namespace ModulesLibrary
         {
         }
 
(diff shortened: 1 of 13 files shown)


TaoziZ03 (approved):
Looks good, thanks!

twsouthwick:
@joperezr can I get a review of this pr?

joperezr (approved):

PR #516, opened 2024-07-12
Use Try... pattern to register services to minimize duplication

twsouthwick:
If someone calls `.AddSystemWebAdapters()` multiple times in order to add more services to it, it registers the initial services multiple times as well. This is ok, except that the startup filter we use to register middleware will now exist multiple times and will register duplicated features on each invocation


joperezr (commented):
Changes look good, but should we also add a test here that calls this methods multiple times and then ensures that the serviceCollection has the right number of services registered?

twsouthwick:
Added a test

joperezr (approved):

PR #523, opened 2024-07-25
Make IRemoteAppAuthenticationService public for use in testing

twsouthwick:
This change also passes the AuthenticationScheme into the AuthenticateAsync method so that the implementations of this interface can be a singleton.

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/Authentication/RemoteAppAuthenticationService.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/Authentication/RemoteAppAuthenticationService.cs
@@ -25,70 +25,51 @@ internal partial class RemoteAppAuthenticationService : IRemoteAppAuthentication
     private readonly HttpClient _client;
     private readonly IAuthenticationResultFactory _resultFactory;
     private readonly ILogger<RemoteAppAuthenticationService> _logger;
-    private readonly IOptionsSnapshot<RemoteAppAuthenticationClientOptions> _authOptionsSnapshot;
-
-    private RemoteAppAuthenticationClientOptions? _options;
+    private readonly IOptionsMonitor<RemoteAppAuthenticationClientOptions> _authOptionsMonitor;
 
     public RemoteAppAuthenticationService(
         IAuthenticationResultFactory resultFactory,
-        IOptionsSnapshot<RemoteAppAuthenticationClientOptions> authOptions,
+        IOptionsMonitor<RemoteAppAuthenticationClientOptions> authOptions,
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/Authentication/RemoteAppAuthenticationService.cs
```diff
@@ -25,70 +25,51 @@ internal partial class RemoteAppAuthenticationService : IRemoteAppAuthentication
...
-    /// An authenticaiton result including the claims principal of the authenticated user (if any)
+    /// An authentication result including the claims principal of the authenticated user (if any)
     /// and a status code and headers to include in the response to the request in case the user could not be authenticated.
     /// </returns>
     /// <exception cref="InvalidOperationException">
     /// An invalid operation exception is thrown if
     /// AuthenticateAsync is called without calling Initialize.
     /// </exception>
-    public async Task<RemoteAppAuthenticationResult> AuthenticateAsync(HttpRequest originalRequest, CancellationToken cancellationToken)
+    public async Task<RemoteAppAuthenticationResult> AuthenticateAsync(AuthenticationScheme scheme, HttpRequest request, CancellationToken cancellationToken)
```
mjrousos, line 56: Nit-pick - let's add an xmldoc param comment for the `scheme` parameter.

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/Authentication/RemoteAppAuthenticationService.cs
```diff
@@ -25,70 +25,51 @@ internal partial class RemoteAppAuthenticationService : IRemoteAppAuthentication
...
     {
-        if (_options is null)
-        {
-            LogHandlerNotInitialized();
-            throw new InvalidOperationException("Remote authentication handler must be initialized before authenticating");
-        }
+        var options = _authOptionsMonitor.Get(scheme.Name);
```
mjrousos, line 58: We should probably have some error checking on options. I could imagine options not including the requested scheme.

mjrousos (approved):
Looks good. A couple minor comments inline.

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/Authentication/RemoteAppAuthenticationService.cs
twsouthwick: an object will always be returned. however, the fields may not be supplied. however, we're using data annotations on this type so it'll check that for us

Issue #526, opened 2024-07-30
Set-Cookie header is not forwarded from authentication server during remote authentication


AndrewBennet:
#### Remote authentication does not forward any Set-Cookie headers from the authentication server back to the response for the initiating request.

My ASP.NET Framework web app has cookie authentication with a cookie authentication provider which periodically re-issues authentication cookies (when a incoming request is valid but is more than a certain proportion of the way through the cookie lifetime). The .NET Framework app's responses thus include a Set-Cookie header with a new authorisation cookie when these conditions are met.

When making a request to the .NET Core app, authorisation is delegated to the .NET Framework app, which correctly returns the serialised user identity claims so the .NET Core app knows the user's identity. But the Set-Cookie header is not copied over from the .NET Framework app response. As a result, if a client only makes requests to the .NET Core app, their login session will never be extended (via issue of a new cookie), until their cookie expires.

### Exceptions (if any)

N/A

### Further technical details 

Upon debugging, I can see that `RemoteAppAuthenticationAuthHandler.HandleAuthenticateAsync()` does not perform the response header propagation, unlike `HandleChallengeAsync(AuthenticationProperties properties)`. The comment in `HandleChallengeAsync` says:

```
// Propagate headers and status code back to the caller
// Different authentication schemes may challenge in different ways in the remote
// app, so make a best effort to forward the effects of these challenges by forwarding
// configured headers (like Location, perhaps) and status code (like 302 or 401, for example).
```

Clearly the Location header is necessary for propagation back to the initiating request when challenging. But note that the `RemoteAppAuthenticationResultFactory` appears to be written to expect header propagation for the authenticate responses too: `RemoteAppAuthenticationClientOptions.ResponseHeadersToForward` is documented as:

```
/// <summary>
/// Gets or sets a list of request headers that should be forwarded to the remote app for authentication purposes. If no headers
/// are specified, all headers will be forwarded.
/// </summary>
```

and its default value includes `Set-Cookie`, which I think means it was expected that the act of authentication should be able to reissue the cookie.

Should these headers be just copied over for authentication calls too, or should there be separate options properties for the sets of headers that should be copied for challenge vs authenticate?

**Other details:**
*ASP.NET Framework Application*:
- Technologies and versions used (i.e. MVC/WebForms/etc):
- .NET Framework Version: 4.7.2
- IIS Version: 10.0.22621.1
- Windows Version: 11 Pro, 23H2

*ASP.NET Core Application*:
[...]

PR #527, opened 2024-07-30
Forward headers received from remote auth call

AndrewBennet:
**PR Title**
Propagate headers from remote authentication service during authentication
Addresses #526

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/Authentication/RemoteAppAuthenticationAuthHandler.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/Authentication/RemoteAppAuthenticationAuthHandler.cs
@@ -67,6 +67,13 @@ internal partial class RemoteAppAuthenticationAuthHandler : AuthenticationHandle
     {
         var authResult = await GetRemoteAppAuthenticationResultAsync();
 
+        // Propagate headers back to the caller. Authentication may reissue an authentication cookie, for instance, so the headers need to be returned
+        // to the caller to ensure they are not lost.
+        foreach (var header in authResult.ResponseHeaders)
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/Authentication/RemoteAppAuthenticationAuthHandler.cs
```diff
@@ -65,7 +65,14 @@ private async Task<RemoteAppAuthenticationResult> GetRemoteAppAuthenticationResu
...
     {
-        var authResult = await GetRemoteAppAuthenticationResultAsync();
+        var authResult = await GetRemoteAppAuthenticationResultAsync();
+
+        // Propagate headers back to the caller. Authentication may reissue an authentication cookie, for instance, so the headers need to be returned
+        // to the caller to ensure they are not lost.
+        foreach (var header in authResult.ResponseHeaders)
+        {
+            Context.Response.Headers.Append(header.Key, header.Value);
+        }
```
mjrousos, line 75: Since this is the same logic as in `HandleChallengeAsync`, it might be better to move it to `GetRemoteAppAuthenticationResultAsync` (just after applying processors), that way the code doesn't need to be duplicated and just runs once right after getting the result.

mjrousos (changes requested):
Looks good, though it may be better to just put the header forwarding logic in `GetRemoteAppAuthenticationResultAsync` so that it's not duplicated.

AndrewBennet:
@dotnet-policy-service agree company="Digital Science"

mjrousos (approved):
LGTM

twsouthwick (approved):
LGTM

PR #529, opened 2024-08-05
Enforce code style in builds

twsouthwick:
Currently, VS keeps telling users that there are style issues and it blocks actual warnings that are important. This turns on code style so that we keep consistency for style choices. Since this hasn't been on before, this change just turns it on and configures the warnings. However, many of the style analyzers are disabled currently which subsequent prs can address.


joperezr (approved):
Thanks!

PR #530, opened 2024-08-07
Update samples

twsouthwick:
This change does the following:

- Create a .slnLaunch file to easily launch the samples (including the multi-app samples)
- Target .NET 8.0 for samples
- Update YARP to 2.1.0
- Clean up registrations to recommended patterns
- Convert samples to use MSBuild.SDK.SystemWeb
- Directory.Build.[props|targets] for samples can now bring in the arcade sdk for more consistent projects


global.json
```diff
@@ -18,6 +18,7 @@
     }
   },
   "msbuild-sdks": {
+    "MSBuild.SDK.SystemWeb": "4.0.88",
```
joperezr, line 21: How will this get updated when new versions are available? Do we care at all?

joperezr (approved):

samples/Directory.Build.props
```diff
@@ -1,7 +1,53 @@
...
+    <!-- MSBuild.SDK.SystemWeb tries to bring in default versions of the compilers, but that is now handled automatically by .NET 9 SDK -->
+    <ExcludeSDKDefaultPackages>true</ExcludeSDKDefaultPackages>
+
+    <!-- This allows the first build to be correct without any warnings on a new clone of the repo -->
+    <GeneratedBindingRedirectsAction>Overwrite</GeneratedBindingRedirectsAction>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <NoWarn>
+      $(NoWarn);
```
joperezr, line 17: NIT: typically you would instead create an .editorconfig or .ruleset where you define which warnings should trigger.

twsouthwick: I agree, but with this it was quick and dirty to remove warnings and I'll follow up with a proper fix to the code patterns

global.json
twsouthwick: Not sure - probably manually until dependabot can handle global.json

PR #532, opened 2024-08-16
Ensure PreSendHeader event is raised when response is started

### LINKED ISSUE #531: Can't edit response headers in `PreSendRequestHeaders` stage if write in a buffering disabled response
### Describe the bug
Can't edit response headers in `PreSendRequestHeaders` stage if write a buffering disabled response

### To Reproduce
```
// somewhere early
responseBodyFeature.DisableBuffering();

...

HttpApplication.ExecuteHandler((...)=>....)
// write a buffering disabled http response

HttpApplication.PreSendRequestHeaders((...) => response.AddHeader("a", ...); // it will throw here. In framework I still can change headers.
```

### Exceptions (if any)
```
System.InvalidOperationException: 'The response headers cannot be modified because the response has already started.'
   at Microsoft.AspNetCore.HttpSys.Internal.HeaderCollection.ThrowIfReadOnly()
   at Microsoft.AspNetCore.HttpSys.Internal.HeaderCollection.set_Item(String key, StringValues value)
   at Microsoft.AspNetCore.Http.ParsingHelpers.AppendHeaderUnmodified(IHeaderDictionary headers, String key, StringValues values)
   at System.Web.HttpResponse.AppendHeader(String name, String value) in System.Web\HttpResponse.cs:line 269
   at System.Web.HttpResponse.AddHeader(String name, String value) in System.Web\HttpResponse.cs:line 264
......
```

### Further technical details 
I am testing 2.0.0-preview1.24405.2.

twsouthwick:
Previously, we were just raining it when a flush occurred, which would be after headers are made readonly.

Fixes #531

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/Features/HttpApplicationPreSendEventsResponseBodyFeature.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/Features/HttpApplicationPreSendEventsResponseBodyFeature.cs
@@ -7,6 +7,7 @@ using System.IO;
 using System.IO.Pipelines;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Web;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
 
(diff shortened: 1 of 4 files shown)


joperezr (approved):
LGTM

Issue #536, opened 2024-08-21
JsonSessionSerializerOptions keys case sensitive


arnoj:
### Describe the bug

Is there a reason the JSON session serializer options has a case sensitive dictionary of known session keys?

I believe the default for .NET has been case insensitive for a long time. Especially in the state sharing use case this might be confusing why values are `null` if the key isn't exactly the same. If this was chosen on purpose it might be an idea to call this out in documentation for clarity?

https://github.com/dotnet/systemweb-adapters/blob/447194bb2fec008473133fb597958e2e77d21945/src/Microsoft.AspNetCore.SystemWebAdapters.Abstractions/SessionState/Serialization/JsonSessionSerializerOptions.cs#L17

joperezr:
I don't believe there is a specific reason. Would you be willing to contribute a PR with the change along with a Comparer on the options?

PR #540, opened 2024-08-29
Case insensitive session keys JSON serializer

arnoj:
- [x] You've read the [Contributor Guide](../CONTRIBUTING.md) and [Code of Conduct](../CODE-OF-CONDUCT.md).
- [x] You've included unit tests for your change, where applicable.
- [ ] You've included inline docs for your change, where applicable.
- [x] There's an open issue for the PR that you are making. If you'd like to propose a new feature or change, please open an issue to discuss the change or find an existing issue.

**PR Title**

Case insensitive session keys for the JSON serializer.

Addresses #536

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.Abstractions/SessionState/Serialization/JsonSessionSerializerOptions.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.Abstractions/SessionState/Serialization/JsonSessionSerializerOptions.cs
@@ -14,7 +14,7 @@ public class JsonSessionSerializerOptions
     /// <summary>
     /// Gets the mapping of known session keys to types
     /// </summary>
-    public IDictionary<string, Type> KnownKeys { get; } = new Dictionary<string, Type>();
+    public IDictionary<string, Type> KnownKeys { get; } = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase);
 
     /// <summary>
     /// Registers a session key name to be of type <typeparamref name="T"/>
--- a/test/Microsoft.AspNetCore.SystemWebAdapters.CoreServices.Tests/SessionState/Serialization/JsonSessionKeySerializerTests.cs
+++ b/test/Microsoft.AspNetCore.SystemWebAdapters.CoreServices.Tests/SessionState/Serialization/JsonSessionKeySerializerTests.cs
@@ -293,6 +293,27 @@ public class JsonSessionKeySerializerTests
         Assert.Equal(primitive, deserialized);
     }
 
+    [Fact]
+    public void CaseInsensitiveSessionKeys()
+    {
+        // Arrange
+        var key = _fixture.Create<string>();
+
+        var options = new JsonSessionSerializerOptions
+        {
+            KnownKeys =
+            {
+                { key, typeof(Type1) },
+            }
+        };
+
+        // Act
+        var result = options.KnownKeys.ContainsKey(key.ToUpperInvariant());
+
+        // Assert
+        Assert.True(result);
+    }
+
     private sealed class Type1
     {
     }


arnoj:
@dotnet-policy-service agree

src/Microsoft.AspNetCore.SystemWebAdapters.Abstractions/SessionState/Serialization/JsonSessionSerializerOptions.cs
```diff
@@ -14,7 +14,7 @@ public class JsonSessionSerializerOptions
     /// <summary>
     /// Gets the mapping of known session keys to types
     /// </summary>
-    public IDictionary<string, Type> KnownKeys { get; } = new Dictionary<string, Type>();
+    public IDictionary<string, Type> KnownKeys { get; } = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase);
```
twsouthwick, line 17: This is a potentially breaking change and I'd prefer it to be done via an additional option:

```diff
public class JsonSessionSerializerOptions
{
+  public IEqualityComparer<string> KeyComparer { get; set; } = StringComparer.Ordinal;
}
```

It should also check if the `KnownKeys` has been created to throw if the setter is called. I'd also be ok if changing the comparer creates a new dictionary with the new comparer and the contents of the previous.

twsouthwick (changes requested):
Thanks for the fix and the tests. I've left a request to retain the original behavior but allow you to override it.

src/Microsoft.AspNetCore.SystemWebAdapters.Abstractions/SessionState/Serialization/JsonSessionSerializerOptions.cs
twsouthwick: If you can show that the session keys on .NET Framework were case insensitive, then we can change the default since we're already doing a major change for v2.0

arnoj: I understand. Hope this is OK.

AFAIR a lot of collections are case insensitive in .NET framework, but session for sure. This SO [comment](https://stackoverflow.com/questions/1731283/#1731535) has some alleged history. I can't find much about classic ASP anymore, but [this](https://github.com/microsoft/referencesource/blob/51cf7850defa8a17d815b4700b67116e3fa283c2/System.Web/State/SessionStateItemCollection.cs#L58) is the Framework location I believe.

test/Microsoft.AspNetCore.SystemWebAdapters.CoreServices.Tests/SessionState/Serialization/JsonSessionKeySerializerTests.cs
```diff
@@ -293,6 +294,70 @@ public void PrimitiveNullableSerializer(int? primitive)
...
+        var result1 = options.KnownKeys.ContainsKey(key1.ToUpperInvariant());
+        var result2 = options.KnownKeys.ContainsKey(key2);
+
+        // Assert
+        Assert.True(result1);
+        Assert.True(result2);
+    }
+
+    [Fact]
+    public void ChangeSessionKeysComparer()
```
twsouthwick, line 324: this test isn't really testing that the comparer is used, just that the keys are still there. Can you verify that the new comparer is use?

twsouthwick (changes requested):
LGTM, just need a few tweaks to a test

test/Microsoft.AspNetCore.SystemWebAdapters.CoreServices.Tests/SessionState/Serialization/JsonSessionKeySerializerTests.cs
arnoj: Sure. It checks the key case insensitively though? The key (`key2`) upper case is verified it does not, and after the change that it does exists (it was registered lower case). That wouldn't work if it if the comparer didn't change?

How would you like me to check? Check the `Dictionary` itself?

```csharp
if (options.KnownKeys is Dictionary<string, Type> dict)
{
    Assert.Equal(dict.Comparer, StringComparer.OrdinalIgnoreCase);
}
```

twsouthwick: oh I guess I read the change too quickly. lgtm :)

twsouthwick (approved):

PR #550, opened 2024-10-28
Support request streams in classic mode in framework services

### LINKED ISSUE #549: Writes to Remote Session failing because session host application does not support ReadEntityBodyMode.Classic
### Describe the bug
Writes to the session made by a .Net8  application do not persist in the session server hosted by a Classic Framework 4.8.1 Web Forms application

### To Reproduce

I can share source code but I'd like to start by describing the situation and seeing if anyone can spot an obvious issue.  Im hoping someone recognizes the very specific errors I'm getting below.

Mostly my code matches whats in this example anyway, except that Im migrating to MVC/.NET8 rather than Blazor:

https://github.com/dotnet/systemweb-adapters/tree/main/samples/WebFormsToBlazor

My WebForms app is OLD and I suspect its missing some key setup but I can't figure it out.

- I am able to run the Session server on the WebForms side and I have the Client going on the .NET8 side.

- I am able read the session on the .NET8 side but...

<b>The writes are failing and it's not a ReadOnly session</b>

Here is my controller in .Net8:

```
[Session(IsReadOnly = false)]
public class HelloWorldController : Controller
{
	public string Welcome()
	{
		// This retrieves an item that was set in the Session server on the WebForms app.  It works.
		object setOnWebFormsSide = System.Web.HttpContext.Current.Session["SetOnWebFormsSide"];

		// This should set a value in the Session server on the WebForms app.  It doesn't work.
		System.Web.HttpContext.Current.Session["Message"] = "Hello";

		return "Hi";
	}}
```

I got the debug symbols for the systemweb-adapters package to trace what happens after I set a session value on the NET8 side:

Inside RemoteAppSessionStateManager.SetOrReleaseSessionData() the line `_backchannelClient.SendAsync()` returns "StatusCode: 404, ReasonPhrase: 'Not Found'" (which is the result passed up from the nested System.Net.Http.HttpClient.SendAsync() call that it makes):

image

At first, I could see that my ASP.NET app does not accept PUT requests at all.   

I can get it to accept PUT requests by copying the `<handlers>` section of the `web.config` from the example linked above, specifically this line:

`<add name="ExtensionlessUrlHandler-Integrated-4.0" path="*." verb="*" type="System.Web.Handlers.TransferRequestHandler" preCondition="integratedMode,runtimeVersionv4.0" />`

But now the same `_backchannelClient.SendAsync()` call from above returns "StatusCode: 500, ReasonPhrase: 'Internal Server Error', ..."

This happens because inside the method:

**System.Web.HttpRequest.GetInputStream(bool,bool)** 

.. it throws an exception **"Incompatible_with_input_stream"** when it sees the ReadEntityMode is **ReadEntityMode.Classic**:

image

Anyone know what this last thing is about?  

It feels like old Webforms apps couldn't handle buffered requests and maybe I need to do some extra setup step to make it support this?  Or I have pipelines setup somewhere that are forcing non-buffered streams but Im not sure where to look.  The WebForms app is very large and hosts web pages and WCF services Im trying to migrate over.  There is a lot in the config files and I don't know where to look.

My WebForms app did not have either the Microsoft.AspNet.Web.Optimization.WebForms or Microsoft.Web.Infrastructure packages like the example but adding these doesn't change the situation.

Any ideas on where to start?

### Exceptions (if any)

See the HttpException thrown in the last screenshot.  I think this is from the HttpRequest code inside the WebForms application.

### Further technical details 

- Technologies and versions used (i.e. MVC/WebForms/etc): WebForms/WCF
- .NET Framework Version: 4.8.1 and .Net Standard 2.0)
- IIS Version:  IIS Express (WebForm application was originally IIS so there could be unknown cruft).   Its also hosted as https on development machines.
- Windows Version: 10

*ASP.NET Core Application*:
- Targeted .NET version: 8
- .NET SDK version: 8.0.400

twsouthwick:
Fixes #549

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/RequestExtensions.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/RequestExtensions.cs
@@ -1,0 +1,37 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Web;
+
+namespace Microsoft.AspNetCore.SystemWebAdapters;
+
+internal static class RequestExtensions
+{
+    /// <summary>
+    /// Gets the request stream. If it has already been retrieved, it will be in whatever <see cref="HttpRequest.ReadEntityBodyMode"/>
+    /// it was already accessed. Otherwise, it will get a bufferless version via <see cref="HttpRequest.GetBufferlessInputStream()"/>.
+    /// </summary>
+    /// <param name="request">The request</param>
+    /// <returns></returns>
+    public static Stream GetInputStream(this HttpRequestBase request)
+    {
+        // If a call has been made to InputStream, we must use that. It is seekable, so we need to rewind it to the beginning
+        if (request.ReadEntityBodyMode == ReadEntityBodyMode.Classic)
+        {
+            var stream = request.InputStream;
+            stream.Position = 0;
+            return stream;
+        }
+
+        // If a call has been made to GetBufferedInputStream, we must use that. It is seekable, so we need to rewind it to the beginning
+        if (request.ReadEntityBodyMode == ReadEntityBodyMode.Buffered)
+        {
+            var stream = request.GetBufferedInputStream();
+            stream.Position = 0;
+            return stream;
+        }
+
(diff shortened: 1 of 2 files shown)


joperezr (approved):
LGTM. should we add any tests here?

PR #551, opened 2024-11-04
Ensure HttpApplication raises SessionStart

twsouthwick:
This fixes two issues that combined do not raise this event:

- A typo sent the event SessionEnd instead of SessionStart
- Wrapped ASP.NET Core session didn't implement the IsNewSession - it was always false. Now we check if there are any keys to determine if it is a new session


mjrousos (approved):
LGTM

PR #553, opened 2024-11-06
Remove public constructor from HttpSessionState

twsouthwick:
There was no public constructor for this on framework. This was added initially as a way to override the current session implementation. However, now it can be changed by setting `HttpContext.Features.Get<ISessionStateFeature>().State` and it will be propagated in the session. This API is no longer needed and not how we want people to use it.

This is a breaking change, but we're already doing that for a few other APIs for v2. This will help guide people to the recommended approach to override behavior.


twsouthwick:
@joperezr this is not a required one, but figured with some of the other API cleanup I did as part of v2, I'd like to see this removed since it's not the recommended path to handle this.

mjrousos (approved):

PR #559, opened 2024-12-04
Copy the keys from framework session state

### LINKED ISSUE #556: Intermittent Error with Remote Session State Sharing in .NET Core 6.0 Application Using SystemWebAdapters
I am reaching out regarding an intermittent issue we are experiencing with session state sharing between our .NET Framework 4.7.2 and .NET Core 6.0 applications using the side-by-side incremental migration approach and the YARP proxy. I am utilizing the System.WebAdapters package to share sessions between the old and new applications, and while this setup generally works as expected, we encounter a sporadic error when accessing certain upgraded pages in our .NET Core 6.0 application.

This error does not occur every time; it seems to happen randomly when old application page stay ideal for around 3-4 min and trying to access upgraded pages in the .NET Core application. And there is traffic on the site of around 10 user accessing the site.

To provide further context:
We are using CustomSerialization to manage dynamic sessions across both applications, which includes serialization and deserialization of session keys in use.
Session sharing is configured to work between the applications and generally behaves correctly, with data persisting accurately in both contexts.

The specific error we receive is as follows:

Unable to load remote session state for session r5fnl1cnxcv53m3okxorxs5y
System.Net.Http.HttpRequestException: Response status code does not indicate success: 302 (Found).
at System.Net.Http.HttpResponseMessage.EnsureSuccessStatusCode()
at Microsoft.AspNetCore.SystemWebAdapters.SessionState.RemoteSession.RemoteAppSessionStateManager.GetSessionDataAsync(String sessionId, Boolean readOnly, HttpContext callingContext, CancellationToken token)
at Microsoft.AspNetCore.SystemWebAdapters.SessionState.RemoteSession.RemoteAppSessionStateManager.CreateAsync(HttpContext context, SessionAttribute metadata)

Please find attached:

Program.cs file from the .NET Core 6.0 application.
global.asax file from the .NET Framework 4.7.2 application.
The CustomSerialization class we are using for session serialization and deserialization in both old and new application.
[CustomSerialization_cs.txt](https://github.com/user-attachments/files/17771166/CustomSerialization_cs.txt)
[Program_cs.txt](https://github.com/user-attachments/files/17771167/Program_cs.txt)
[Global_asax_cs.txt](https://github.com/user-attachments/files/17771165/Global_asax_cs.txt)

Any insights into the cause of this intermittent error, and guidance on how to ensure consistent session state sharing, would be greatly appreciated. 

- Technologies and versions used (i.e. MVC/WebForms/etc):
- IIS Version: 10
- Windows Version: Windows Server 2016

*ASP.NET Core Application*:
- Targeted .NET version: .Net Core 6.0

Thank you in advance for your assistance.

twsouthwick:
It appears that some session providers on the framework side may not be able to iterate over the keys while grabbing items as it may throw an InvalidOperationException with the error: Collection was modified after the enumerator was instantiated. Even though it appears that this should not happen, we'll copy the keys over before grabbing items from the collection so this can't happen.

Fixes #556

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/SessionState/Serialization/HttpSessionStateBaseWrapper.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/SessionState/Serialization/HttpSessionStateBaseWrapper.cs
@@ -2,12 +2,10 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Collections.Generic;
-using System.Collections.Specialized;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Web;
-using System.Web.SessionState;
 
 namespace Microsoft.AspNetCore.SystemWebAdapters.SessionState;
 
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/SessionState/Serialization/HttpSessionStateBaseWrapper.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/SessionState/Serialization/HttpSessionStateBaseWrapper.cs
@@ -2,10 +2,12 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Collections.Generic;
+using System.Collections.Specialized;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Web;
+using System.Web.SessionState;
 
 namespace Microsoft.AspNetCore.SystemWebAdapters.SessionState;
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/SessionState/Serialization/HttpSessionStateBaseWrapper.cs
```diff
@@ -2,10 +2,12 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Collections.Generic;
+using System.Collections.Specialized;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Web;
+using System.Web.SessionState;
```
mjrousos, line 10: These new `using` statements look unnecessary.

mjrousos (approved):
Left a minor comment about some unnecessary using statements. Looks good otherwise!

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/SessionState/Serialization/HttpSessionStateBaseWrapper.cs
twsouthwick: good catch

PR #561, opened 2024-12-06
Use HTTP2 full duplex for committing session

twsouthwick:
The remote session sharing current creates two connections when in non-readonly mode:

- One request to GET the session data
- One request to PUT the session data

This PR adds a POST verb to the endpoint that expects full-duplex streaming to be available and allows
for receiving the session state first and then sending the result back once committed.

Since this requires HTTP2 and SSL, there is a flag to go back to the original implementation as well.

For readonly mode, the original client implementation is used as it calls the GET endpoint and doesn't
need to send any results back.


src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/SessionState/RemoteSession/RemoteAppSessionDispatcher.cs
```diff
@@ -0,0 +1,107 @@
...
+/// This is used to dispatch to either <see cref="DoubleConnectionRemoteAppSessionManager"/> if in read-only mode or <see cref="SingleConnectionWriteableRemoteAppSessionStateManager"/> if not.
+/// </summary>
+internal sealed partial class RemoteAppSessionDispatcher : ISessionManager
+{
+    private readonly IOptions<RemoteAppSessionStateClientOptions> _options;
+    private readonly ISessionManager _singleConnection;
+    private readonly ISessionManager _doubleConnection;
+    private readonly ILogger _logger;
+
+    public static ISessionManager Create(
```
mjrousos, line 24: Wouldn't we always just get this from DI? Why do we need a static create method?

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/SessionState/RemoteSession/RemoteAppSessionStateExtensions.cs
```diff
@@ -13,10 +14,20 @@ public static ISystemWebAdapterRemoteClientAppBuilder AddSessionClient(this ISys
...
+        builder.Services.AddTransient<DoubleConnectionRemoteAppSessionManager>();
+        builder.Services.AddTransient<SingleConnectionWriteableRemoteAppSessionStateManager>();
+        builder.Services.AddTransient<ISessionManager, RemoteAppSessionDispatcher>();
 
         builder.Services.AddOptions<RemoteAppSessionStateClientOptions>()
             .Configure(configure ?? (_ => { }))
+            .PostConfigure<IOptions<RemoteAppClientOptions>>((options, remote) =>
+            {
+                // The single connection remote app session client requires https to work so if that's not the case, we'll disable it
+                if (!string.Equals(remote.Value.RemoteAppUrl.Scheme, "https", StringComparison.OrdinalIgnoreCase))
```
mjrousos, line 26: Might be worth logging a warning here to let the user know why single connection isn't being used.

mjrousos (approved):
LGTM

src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/SessionState/RemoteSession/RemoteAppSessionDispatcher.cs
twsouthwick: For simplicity in testing, I want to be able to just pass in a mock so this facilitates that

twsouthwick: also added a comment to the method to make it clear what it's for

PR #566, opened 2024-12-11
Build samples on ci

twsouthwick:


samples/targets/WebApplications/Microsoft.WebApplication.targets
```diff
@@ -0,0 +1,271 @@
+
```
twsouthwick, line 1: @joperezr this is a targets file that is normally shipped with VS and the only blocker now for building the samples on CI. I do not know its license. Do you know how we can figure out if it's something we can distribute? If so, I'd want to push it upstream to the MSBuild.SDK.SystemWeb project after we get things working here.

PR #568, opened 2024-12-16
Support serializing session changesets

### LINKED ISSUE #503: Session Values are set to Null if you dont map them
### Describe the bug
When using ThrowOnUnknownSessionKey = false; with the options, it allows you to ignore keys that aren't yet mapped.  However, those session values are nulled out rather than ignored.  I'm not sure if this is intentional, but when converting large sites a page at a time, it would make sense if there was a way to completely ignore these keys that you don't need yet.

### To Reproduce

.NET Framework Side:

Configure to not throw on unknown keys and only register one key but then set 2 different keys in session:
image
image

.NET 8 Side:

Configure to not throw on unknown keys and only register one key, view both keys and note that only 1 is set:
image
image

Load the .NET Framework side again and note test2 is now null:
image

### Further technical details 

- MVC
- IIS Version: 
- Windows Version: 11

*ASP.NET Core Application*:
- Targeted .NET version: .NET 8
- .NET SDK version:

twsouthwick:
Previously, the serialization of the session state to go between the remote server and the client would
deserialize/serialize every key even if it hadn't changed. As part of this change:

- The session state tracks which items are new or have been accessed (and thus potentially changed)
- Only sends a diff list rather than all the items
- Only deserializes items if they are being accessed
- Makes ISessionSerializer internal since we expect people to extend via ISessionKeySerializer instead

Fixes #503

### DIFF
--- a/test/Microsoft.AspNetCore.SystemWebAdapters.CoreServices.Tests/SessionState/Serialization/BinarySessionSerializerTests.cs
+++ b/test/Microsoft.AspNetCore.SystemWebAdapters.CoreServices.Tests/SessionState/Serialization/BinarySessionSerializerTests.cs
@@ -11,6 +11,7 @@ using Xunit;
 
 namespace Microsoft.AspNetCore.SystemWebAdapters.SessionState.Serialization.Tests;
 
+[System.Diagnostics.CodeAnalysis.SuppressMessage("Design", "CA1062:Validate arguments of public methods", Justification = "Test arguments")]
 public class BinarySessionSerializerTests
 {
     [InlineData(new byte[] { 1, 2, 105, 100, 0, 0, 0, 0, 0, 0 })]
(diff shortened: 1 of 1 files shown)


src/Services/SessionState/SessionStateCollection.cs
```diff
@@ -0,0 +1,215 @@
...
+    public ISessionKeySerializer Serializer { get; }
+
+    public void AddUnknownKey(string key)
+    {
+        (UnknownKeys ??= new()).Add(key);
+    }
+
+    public void MarkUnchanged(string key) => _items[key] = new([]);
+
+    public void MarkRemoved(string key) => _items[key] = new((object?)null);
```
mjrousos, line 54: Won't this cause the item to appear as new since it's using ItemHolder's object ctor?

mjrousos (commented):
Looks good. I left a few questions inline.

src/Services/SessionState/SessionStateCollection.cs
twsouthwick: Yeah. It's not a real problem because of how that property is currently used, but isn't acurate. Adding some tests for that class before committing to verify behavior

designs/session-serialization.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+## Full Copy (Mode = 1)
+
+The following is the structure of the key blobs when the full state is serialized:
+
+```mermaid
+packet-beta
+0-10: "Key name"
+11-20: "Serialized value"
```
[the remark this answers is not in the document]
twsouthwick: Unfortunately, there is no way to turn that off for now. The packet mermaid stuff is in beta, so I'll look to see if there's a way to request that

twsouthwick:
This breaks with the new http2 full duplex mode. will need to rework some things.

For now, to not block getting v2.0 of the adapters out, we'll just get #573 in

PR #574, opened 2025-01-15
Consolidate Directory.Packages.props and update to .NET 8 libraries

twsouthwick:
This brings all the dependencies of the project into a single place so that we can more easily update them. As part of this, the Microsoft.Extensions.* libraries are updated to the LTS .NET 8 version. These are only really surfaced to the end user as part of what is included in the FrameworkServices package since the .NET 6/8 builds just reference the ASP.NET Core framewrok.


joperezr (approved):
Thanks @twsouthwick!

PR #575, opened 2025-01-15
Drop .NET 6 runtime support (compiling still works for libraries)

twsouthwick:


joperezr (approved):
Assuming build passes, this LGTM. THanks so much for doing this @twsouthwick

PR #596, opened 2025-06-12
Use IHostApplicationBuilder for framework application to enable Aspire development

twsouthwick:
This change adds a new way to configure the framework application for use with incremental migration that unlocks a lot of goodness from Aspire.

Here's an example of traces showing the incremental migration steps:

Screenshot 2025-06-11 164411

Here's a screen shot of the metrics gathered on the framework application:

image

This is enabled by the following:

- An IHostApplicationBuilder-derived HttpApplicationHostBuilder following the pattern of the WebApplicationBuilder
- An IHost-derived HttpApplicationHost following the pattern of WebApplication

Once these are available (and hooked up), then a lot of the APIs for the Aspire-related ServiceDefaults becomes available. To see how the samples make use of this, see: https://github.com/dotnet/systemweb-adapters/blob/480f19928fc78e4d69e8056b03283dea1f490bd9/samples/ServiceDefaults/Extensions.cs#L22-L141


twsouthwick:
@CZEMacLeod I'd appreciate any of your thoughts here. It was partially inspired by some of your work - but wanted it minimize any specialized overloads that need to be configured to keep it pretty open

CZEMacLeod:
This all looks pretty good. Will have to try it out and see how it compares to my implementation, and if my scope per request, webapi and mvc dependency injection providers can hook up easily to this.

I feel like `IHost` and (or at least) `IServiceProvider` could be added to the application state when the host starts.
Also perhaps an extension method for GetApplicationServices() => HttpApplicationHost.Current.Services on httpcontext?
It's a pity that the new c# 14 extension properties won't work for net48 - as it would be lovely to have `ApplicationServices` and `RequestServices` to match the aspnet core versions.

I don't think you need to have the abstract check in WebObjectActivator - the inbuilt code doesn't do the check when WOA is not set - it would just throw. I would also special case `IServiceProvider` and `IKeyedServiceProvider`, so they don't resolve through the serviceprovider but just return it.

```cs
    public object? GetService(Type serviceType)
    {
        if (serviceType == typeof(IServiceProvider))
        {
            return (object?)serviceProvider;
        }
        if (serviceType == typeof(IKeyedServiceProvider))
        {
            return serviceProvider as IKeyedServiceProvider ?? serviceProvider.GetService(serviceType);
        }
        return serviceProvider.GetService(serviceType) ??
            // The implementation of dependency injection in System.Web expects to be able to create instances
            // of non-public and unregistered types.
            Activator.CreateInstance(
            serviceType,
            BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.CreateInstance,
            null,
            null,
            null);
    }
```

```cs
public static partial class HttpApplicationDependencyInjectionExtensions
{
#if NET48_OR_GREATER
    public readonly static string ServiceProviderKey = typeof(IServiceProvider).FullName!;

    private static IServiceProvider? GetDefaultServiceProvider() =>
        HttpRuntime.WebObjectActivator?.GetService<IServiceProvider>();

    public static IServiceProvider? GetServiceProvider(this HttpApplication application) =>
        application.Application.GetServiceProvider() ?? GetDefaultServiceProvider();

    private static IServiceProvider? GetServiceProvider(this HttpApplicationState state) =>
        state[ServiceProviderKey] as IServiceProvider ?? GetDefaultServiceProvider();

    public static void SetServiceProvider(this HttpApplication application, IServiceProvider serviceProvider) =>
        application.Application.SetServiceProvider(serviceProvider);

    public static void SetServiceProvider(this HttpApplicationState state, IServiceProvider serviceProvider)
    {
        if (serviceProvider is null)
        {
            throw new ArgumentNullException(nameof(serviceProvider));
        }

        state[ServiceProviderKey] = serviceProvider;
        HttpRuntime.WebObjectActivator = new WebObjectActivator(serviceProvider);
    }

    public static void ClearServiceProvider(this HttpApplicationState state)
    {
        state.Remove(ServiceProviderKey);
        HttpRuntime.WebObjectActivator = null;
    }
#endif
#if NET8_0_OR_GREATER
    public static IServiceProvider? GetServiceProvider(this HttpApplication application) => 
        HttpRuntime.WebObjectActivator;
#endif
}
```

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/SystemWebAdapterExtensions.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        builder.Services.AddHostedService<RegisteredHostServices>();
+
+        return builder;
+    }
+
+    public static ISystemWebAdapterBuilder AddSystemAdapters(this IServiceCollection services)
+       => new SystemWebAdapterBuilder(services);
+
+    private sealed class RegisteredHostServices : IServiceProvider, IHostedService
```
CZEMacLeod, line 30: This name feels odd to me - maybe WebObjectActivatorHostService?

CZEMacLeod:
In your `HttpApplicationHostBuilder`, you might want to copy the web.config settings into the IConfiguration - either automatically, or provide an easy mechanism for it.
```cs
using Microsoft.Extensions.Configuration;
using System;
using System.Collections.Generic;
using System.Collections.Specialized;
using System.Configuration;
using System.Linq;

namespace C3D.Extensions.Configuration;

public static class ConfigurationBuilderNameValueExtensions
{
    public static IConfigurationBuilder AddNameValueCollection(this IConfigurationBuilder builder,
        NameValueCollection collection) =>
            builder.AddInMemoryCollection(
                collection.AllKeys.Select(key => new KeyValuePair<string, string?>(key, collection[key])));

    public static IConfigurationBuilder AddAppSettings(this IConfigurationBuilder builder) =>
        builder
            .AddNameValueCollection(System.Configuration.ConfigurationManager.AppSettings);

    public static IConfigurationBuilder AddConnectionStrings(this IConfigurationBuilder builder,
        string? providerFilter = null,
        Func<ConnectionStringSettings, string>? extractConnectionString = null)
    {
        var connections = System.Configuration.ConfigurationManager.ConnectionStrings;

        extractConnectionString ??= static cnn => cnn.ConnectionString;

        return builder.AddInMemoryCollection(
                connections
                    .OfType<ConnectionStringSettings>()
                    .Where(cnn => providerFilter is null || cnn.ProviderName == providerFilter)
                    .Select(cnn => new KeyValuePair<string, string?>("ConnectionStrings:" + cnn.Name, extractConnectionString(cnn)))
                );
    }
}
```

src/Microsoft.AspNetCore.SystemWebAdapters/Hosting/HostingEnvironment.cs
```diff
@@ -1,56 +1,56 @@
-// Licensed to the .NET Foundation under one or more agreements.
```
CZEMacLeod, line 1: What changed in this file that made every line but 1 show a change with no actual changes?

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/Hosting/HttpApplicationHost.cs
```diff
@@ -0,0 +1,53 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
CZEMacLeod, line 1: It would actually be really nice to have this as a separate package, perhaps with some of the DI stuff, and maybe even the ContentFile sample extensions in it, that could be used in an ASP.Net 4.x project without SWA at all.
I guess there isn't a huge overhead of pulling in the whole `Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices` package, but if you simply wanted to use the aspire integration part it might be nice to have.

samples/AuthRemoteFormsAuth/AuthRemoteFormsAuthFramework/Global.asax.cs
```diff
@@ -4,21 +4,27 @@
...
-                .AddVirtualizedContentDirectories()
-                .AddProxySupport(options => options.UseForwardedHeaders = true)
-                .AddRemoteAppServer(options =>
-                    {
-                        options.ApiKey = ConfigurationManager.AppSettings["RemoteApp__ApiKey"];
-                    })
-                .AddAuthenticationServer();
+            HttpApplicationHost.RegisterHost(builder =>
+            {
+                builder.AddServiceDefaults();
```
joperezr, line 19: One other typical thing that is done when you add aspire to an app is to call abb.MapDefaultEndpoints() which would get things like healthchecks working. That would obviously not go here but just mentioning it in case you want to add that as well.

samples/AuthRemoteFormsAuth/AuthRemoteFormsAuthAppHost/AuthRemoteFormsAuthAppHost.csproj
```diff
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net9.0-windows</TargetFramework>
+    <TargetFramework>net10.0-windows</TargetFramework>
```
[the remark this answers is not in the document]
twsouthwick: yeah, I was actually having issues in the repo with targeting net9.0-windows and net10.0-windows seems to work.

PR #602, opened 2025-07-01
Condition out Linux and Mac builds on official builds

joperezr:
Official builds should only build Windows as that is the one that produces all of the signed packages.

cc: @twsouthwick


twsouthwick (approved):
It is weird that this only started failing recently. It's probably fine (to at unblock builds) but seems like it would nice to validate everything when doing the "official" builds

PR #618, opened 2025-10-15
Add a static config accessor that can help with migration

twsouthwick:
This enables switching from `ConfigurationManager.[Appsettings|ConnectionStrings]` to `AppConfiguration.[GetSetting|GetConnectionString]` so that there is a static configuration source on all platforms while adapting to more modern practices of dependency injection for configuration.

### DIFF
--- a/samples/AppConfig/AppConfigFramework/ConfigInfo.cs
+++ b/samples/AppConfig/AppConfigFramework/ConfigInfo.cs
@@ -3,6 +3,7 @@ using System.Linq;
 using System.Text.Json;
 using System.Web;
 using System.Web.SessionState;
+using Microsoft.AspNetCore.SystemWebAdapters;
 
 namespace AppConfigFramework;
 
--- a/samples/AppConfig/AppConfigFramework/Global.asax.cs
+++ b/samples/AppConfig/AppConfigFramework/Global.asax.cs
@@ -1,5 +1,6 @@
 using System.Web;
 using Microsoft.AspNetCore.SystemWebAdapters.Hosting;
+using Microsoft.Extensions.Hosting;
 
 namespace AppConfigFramework
 {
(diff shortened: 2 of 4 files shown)


samples/AppConfig/AppConfigFramework/ConfigInfo.cs
```diff
@@ -0,0 +1,27 @@
...
+    public bool IsReusable => true;
+
+    public void ProcessRequest(HttpContext context)
+    {
+        context.Response.ContentType = "application/json";
+        context.Response.StatusCode = 200;
+
+        context.Response.Write(JsonSerializer.Serialize(new
+        {
+            Setting1 = AppConfiguration.GetSetting("Setting1"),
```
joperezr: Is the plan for this to look the same on Core? Or is the plan to have folks use DI Configuration there?

src/Microsoft.AspNetCore.SystemWebAdapters/AppConfiguration.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Web;
+using Microsoft.AspNetCore.SystemWebAdapters.Adapters;
+
+namespace Microsoft.AspNetCore.SystemWebAdapters;
+
+#if !NET45
+/// <summary>
+/// Provides a static way to access application configuration settings while migrating to Microsoft.Extensions.Configuration types.
+/// </summary>
+public static class AppConfiguration
```
joperezr: Do we want to mark this as experimental or are we happy with its shape?

joperezr (approved):
Left a few minor comments but looks good otherwise

test/Microsoft.AspNetCore.SystemWebAdapters.E2E.Tests/AppConfigTests.cs
```diff
@@ -0,0 +1,34 @@
...
+using Xunit;
+
+namespace Microsoft.AspNetCore.SystemWebAdapters.E2E.Tests;
+
+public class AppConfigTests(AspireFixture<AppConfigAppHost> aspire) : IClassFixture<AspireFixture<AppConfigAppHost>>
+{
+    [Fact]
+    public async Task ConfigurationIsConfigured()
+    {
+        var app = await aspire.GetApplicationAsync();
```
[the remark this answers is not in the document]
twsouthwick: update so they do

src/Microsoft.AspNetCore.SystemWebAdapters/AppConfiguration.cs
twsouthwick: it's a simple enough shape that don't expect to change drastically or in a breaking way

PR #639, opened 2025-11-18
Fix issue when enabling static user accessors

twsouthwick:
The RequestUserFeature implements the ASP.NET Core IHttpAuthenticationFeature as well as the adapter's IRequestUserFeature. We were only copying the user to the static accessors when the ASP.NET Core User was set not when the IRequestUserFeature one was. Since there's a single User property that stores it for both, this change moves the call to sync to the static accessors there.

### DIFF
--- a/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/Middleware/RequestUserFeaturesMiddleware.cs
+++ b/src/Microsoft.AspNetCore.SystemWebAdapters.CoreServices/Middleware/RequestUserFeaturesMiddleware.cs
@@ -53,6 +53,9 @@ internal sealed partial class RequestUserFeaturesMiddleware(RequestDelegate next
             set
             {
                 field = value;
+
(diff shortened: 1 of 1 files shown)


test/Microsoft.AspNetCore.SystemWebAdapters.CoreServices.Tests/MachineKeyTests.cs
```diff
@@ -86,7 +87,7 @@ public void UnprotectWithPurposes(params string[] purposes)
         Assert.Equal(plaintext, result);
     }
 
-    private static IDataProtectionProvider BuildProvider(params string[] purposes)
+    private static IDataProtectionProvider BuildProvider(params IEnumerable<string> purposes)
```
twsouthwick: this is due to some overload resolution changes with c# version change. pretty cool to use a params with IEnumerable<> though

PR #672, opened 2026-04-07
Bring Microsoft.AspNetCore.DataProtection.SystemWeb compatibility mode into the FrameworkServices project 

twsouthwick:


src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/SystemWebDataProtectionExtensions.cs
```diff
@@ -4,78 +4,105 @@
...
-        }
+        TrySetMachineKeyOverride();
 
-        if (!TrySetStartupType())
-        {
-            throw new InvalidOperationException($"Must not manually set the '{StartupTypeKey}' app setting when using HttpApplicationHostBuilder.AddDataProtection.");
-        }
+        builder.Services.TryAddSingleton<IApplicationDiscriminator, SystemWebApplicationDiscriminator>();
 
         return builder.Services.AddDataProtection();
```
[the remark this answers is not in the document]
twsouthwick: e2e testing exists already to verify this gets hooked up

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/SystemWebDataProtectionExtensions.cs
```diff
@@ -4,78 +4,105 @@
...
+    private static void TrySetMachineKeyOverride()
+    {
+        if (_isSet)
         {
-            return false;
+            return;
         }
 
-        ConfigurationManager.AppSettings[StartupTypeKey] = startupType;
+        _isSet = true;
```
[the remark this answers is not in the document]
twsouthwick: @copilot apply changes based on [this feedback](https://github.com/dotnet/systemweb-adapters/pull/672#discussion_r3048383298)

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/MachineKeyExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Reflection;
+using System.Web.Configuration;
+
+namespace System.Web;
+
+internal static class MachineKeyExtensions
+{
+
```
twsouthwick:
```suggestion
```

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/SystemWebDataProtectionExtensions.cs
```diff
@@ -4,78 +4,113 @@
...
+            // at the web app root. If the value was set explicitly (even if the value is empty),
+            // honor it as the discriminator.
+            var machineKeySection = (MachineKeySection)WebConfigurationManager.GetWebApplicationSection("system.web/machineKey");
+            if (machineKeySection.ElementInformation.Properties["applicationName"].ValueOrigin != PropertyValueOrigin.Default)
+            {
+                return machineKeySection.ApplicationName;
+            }
+            else
+            {
+                return env.ApplicationName;
```
CZEMacLeod: The previous implementation used `HttpRuntime.AppDomainAppId` and the default aspnetcore implementation uses ContentRootPath instead of ApplicationName, which by default is the main assembly name (which can cause some problems as some bits of code assume this is an actual assembly, where others think it is an arbitrary, user settable string).
https://github.com/dotnet/dotnet/blob/8fe340e65f698e5f29a8357f1b0a135884694c3e/src/aspnetcore/src/DataProtection/DataProtection/src/Internal/HostingApplicationDiscriminator.cs#L30-L43
https://github.com/dotnet/aspnetcore/blob/acdbcb0cbd0ce2470713c44393895dc8a6ad462a/src/DataProtection/SystemWeb/src/DataProtectionStartup.cs#L83-L101

It would be an extra dependency, but consider using `Microsoft.Extensions.AmbientMetadata.Application` and getting `IOptions<ApplicationMetadata>.Value.ApplicationName` instead of `IHostEnvironment.ApplicationName` when situations like this occur.

I think the reason for using `HttpRuntime.AppDomainAppId` or `ContentRootPath` is such that two instances of the same site, do not (by default) use the same encryption key.

CZEMacLeod:
Just as a comment - Despite using reflection to do it, I love that this does not require modifying `web.config` to setup the machine key.

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/SystemWebDataProtectionExtensions.cs
twsouthwick: good point. I went with `IHostEnvironment.ApplicationName` as it can be overridden however someone wants. Right now it defaults to `HostingEnvironment.SiteName`, which in my testing is the same as `HttpRuntime.AppDomainAppId`. Since the default is relatively new, I may switch the default app name to `HttpRuntime.AppDomainAppId` so that it aligns better here. Thoughts?

twsouthwick: I think I initially missed the issue you're rasing:

> default aspnetcore implementation uses ContentRootPath instead of ApplicationName, 

hmm this makes things a bit tricky. I think it would be better to go back to what the default has been and make sure to update the docs that you should manually be calling `.SetApplicationName(name)`

CZEMacLeod: I wasn't strictly aware that you initialized IHostEnvironment.ApplicationName to HostingEnvironment.SiteName.
That seems likely to break UserSecrets configuration based on https://github.com/dotnet/runtime/issues/114505
There are some other similar issues raised about the fact that IHostEnvrironment.ApplicationName should be the entry point assembly name, and not set to an arbitrary string.

I think an extension method to set the dataprotection application discriminator from the machinekey section (or default to the original implementation) would be one way to go.

Simply registering an `IConfigureOptions<Microsoft.AspNetCore.DataProtection.DataProtectionOptions>` _after_ the call to `builder.Services.AddDataProtection()` allows you to (conditionally) override the `ApplicationDiscriminator`.
This would be called after the default implementation in https://github.com/dotnet/dotnet/blob/main/src/aspnetcore/src/DataProtection/DataProtection/src/Internal/DataProtectionOptionsSetup.cs which uses `IApplicationDiscriminator`. You would need to replicate the small bit of logic in 
https://github.com/dotnet/dotnet/blob/e52493553982a80cd4c7d372c104e3a9dbff0e0d/src/aspnetcore/src/DataProtection/DataProtection/src/DataProtectionUtilityExtensions.cs#L44-L46

```cs
namespace Microsoft.AspNetCore.DataProtection;

using Microsoft.Extensions.DependencyInjection;
using System.Configuration;
using System.Web.Configuration;

public static class DataProtectionExtensions
{
    public static IDataProtectionBuilder UseMachineKeyApplicationNameDiscriminator(this IDataProtectionBuilder builder)
    {
        builder.Services.Configure<DataProtectionOptions>(options =>
        {
            // Try reading the discriminator from <machineKey applicationName="..." /> defined
            // at the web app root. If the value was set explicitly (even if the value is empty),
            // honor it as the discriminator.
            var machineKeySection = (MachineKeySection)WebConfigurationManager.GetWebApplicationSection("system.web/machineKey");
            if (machineKeySection.ElementInformation.Properties["applicationName"].ValueOrigin != PropertyValueOrigin.Default)
            {
                var applicationName = machineKeySection.ApplicationName?.Trim();
                options.ApplicationDiscriminator = string.IsNullOrEmpty(applicationName) ? null : applicationName;
            }
        });
        return builder;
    }

    public static IDataProtectionBuilder UseAppDomainAppIdDiscriminator(this IDataProtectionBuilder builder)
    {
        builder.Services.Configure<DataProtectionOptions>(options =>
        {
            options.ApplicationDiscriminator = System.Web.HttpRuntime.AppDomainAppId.Trim();
        });
        return builder;
    }
}
```

Alternatively, you could bake in the machineKey check in 
https://github.com/dotnet/systemweb-adapters/blob/f3929755286d2e359fc31de83e8cab509f054139/src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/SystemWebDataProtectionExtensions.cs#L46
as
```cs
        return builder.Services.AddDataProtection(options =>
        {
            // Try reading the discriminator from <machineKey applicationName="..." /> defined
            // at the web app root. If the value was set explicitly (even if the value is empty),
            // honor it as the discriminator.
            var machineKeySection = (MachineKeySection)WebConfigurationManager.GetWebApplicationSection("system.web/machineKey");
            if (machineKeySection.ElementInformation.Properties["applicationName"].ValueOrigin != PropertyValueOrigin.Default)
            {
                var applicationName = machineKeySection.ApplicationName?.Trim();
                options.ApplicationDiscriminator = string.IsNullOrEmpty(applicationName) ? null : applicationName;
            }
        });
```

An even more configurable version might be
```cs
/// <summary>
/// Adds <see cref="IDataProtectionProvider"/> to the <see cref="HttpApplicationHost"/> and enables <see cref="MachineKey"/> integration.
/// </summary>
/// <param name="builder">The <see cref="HttpApplicationHostBuilder"/>.</param>
public static IDataProtectionBuilder AddDataProtection(this HttpApplicationHostBuilder builder, 
    Action<DataProtectionOptions> setupAction = null,
    bool useMachineKeyApplicationName = true)
{
    if (builder is null)
    {
        throw new ArgumentNullException(nameof(builder));
    }

    if (ConfigurationManager.GetSection("system.web/machineKey") is MachineKeySection section)
    {
        if (!string.IsNullOrEmpty(section.DataProtectorType))
        {
            throw new InvalidOperationException("Could not setup DataProtection for use with MachineKey due to existing configuration of system.web/machineKey");
        }
    }

    // We use this to auto-start it ASAP to ensure the dataprotector is set up before anyone else tries to do anything
    builder.Services.AddHostedService<MachineKeySetup>();

    return builder.Services.AddDataProtection(options =>
    {
        if (useMachineKeyApplicationName)
        {
            // Try reading the discriminator from <machineKey applicationName="..." /> defined
            // at the web app root. If the value was set explicitly (even if the value is empty),
            // honor it as the discriminator.
            var machineKeySection = (MachineKeySection)WebConfigurationManager.GetWebApplicationSection("system.web/machineKey");
            if (machineKeySection.ElementInformation.Properties["applicationName"].ValueOrigin != PropertyValueOrigin.Default)
            {
                var applicationName = machineKeySection.ApplicationName?.Trim();
                options.ApplicationDiscriminator = string.IsNullOrEmpty(applicationName) ? null : applicationName;
            }
        }
        setupAction?.Invoke(options);
    });
}
```

HamsterExAstris:
Thanks for the quick turnaround from the deprecation announcement. :) As current users of the data protection functionality, it looks like this'll work for us.

(We're calling `IDataProtectionBuilder.SetApplicationName()`, so I don't think we have any particular horse on the default ApplicationName logic question above.)

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/SystemWebDataProtectionExtensions.cs
twsouthwick: > I wasn't strictly aware that you initialized IHostEnvironment.ApplicationName to HostingEnvironment.SiteName.
That seems likely to break UserSecrets configuration based on https://github.com/dotnet/runtime/issues/114505
There are some other similar issues raised about the fact that IHostEnvrironment.ApplicationName should be the entry point assembly name, and not set to an arbitrary string.

updated to do a best search for an "entry assembly" in asp.net framework: https://github.com/dotnet/systemweb-adapters/pull/674

twsouthwick: as for the other stuff, I think we should probably keep to what it has been doing, and people can set the application name using `SetApplicationName` extension method

src/Microsoft.AspNetCore.SystemWebAdapters.FrameworkServices/CompatibilityDataProtector.cs
```diff
@@ -0,0 +1,118 @@
...
+            _lazyProtectorSuppressedPrimaryPurpose = new Lazy<IDataProtector>(() => GetDataProtectionProvider().CreateProtector(specificPurposes));
+        }
+        else
+        {
+            _lazyProtectorSuppressedPrimaryPurpose = _lazyProtector;
+        }
+    }
+
+    // We take care of flowing purposes ourselves.
+    protected override bool PrependHashedPurposeToPlaintext { get; }
```
KalleOlaviNiemitalo: I don't see anything setting this property.  Could then be `{ get => false; }` so as not to generate an instance field.

twsouthwick: This is something that existed in the previous version (the 2.3 shim library) so that people could (if they wanted) override the CompatibilityDataProtector and set it themselves manually. Not sure how much we need that behavior, or if we want to rethink a design that fits in with the more IHostApplicationBuilder pattern

twsouthwick: oh I guess your point is still valid if we keep it - I misunderstood initially. Sure I'll change it so it doesn't create a field. Deciding if we want to keep the same pattern is also worth having

PR #667, opened 2026-02-20
Support HostApplicationBuildersSettings for framework application (#665)

EmperorArthur:
This updates HttpApplicationHost.CreateBuilder to more closely match HostApplicationBuilder's constructor.  Supports setting any HostDefaults value via configuration or a HostApplicationBuildersSettings instance. This allows even more patterns used in modern ASP.NET Core applications to be available in ASP.NET Framework, including setting the application name and environment.

PR #674, opened 2026-04-08
Use assembly name of global.asax type as default IHostEnvironment.ApplicationName

twsouthwick:
There are a lot of assumptions that the ApplicationName is actually the assembly name and cause some things (like user secrets) to break if it is not.

Since there's no direct way to get the main assembly directly, this sets it by searching the current HttpApplication as a best effort for this.


EmperorArthur:
I see what you're doing, but am not sure it's the best approach.  Especially since it's still relying on other assumptions.  Ones which I know would fail for my company's sites.

### Alternatives
There are several options to work around the problem that I believe are better alternatives.

1. When the next version is released, a user can set "applicationName" in `GlobalConfiguration.Configure` to whatever it should be.  (This ability was added in PR #667)
2. Make `HttpApplicationHostBuilder.BuildAndRunInBackground` and `HttpApplicationHost.CreateBuilder(HostApplicationBuilderSettings settings)` public
3. Add a 2nd argument/version of `HttpApplicationHost.RegisterHost` that accepts an "HostApplicationBuilderSettings".

As I noted, this is an issue which I have had to deal with, and the 2nd option was probably on my mind when I worked on PR #667.

### Other Info

* The .NET Core extensions [use](https://github.com/dotnet/dotnet/blob/d3d9bbbf2028bc388d245a60a03d216c3582f605/src/aspnetcore/src/Hosting/Hosting/src/WebHostBuilderExtensions.cs#L41) the assembly name of whatever is configuring the builder.  Which is different from both the current and new approach.
* Some of the weird formatting I see cleaned up is because that snippet is based on. [HostApplicationBuilder.Initialize](https://github.com/dotnet/dotnet/blob/a8b33e7593686eaee701cd124daaabff2311634f/src/runtime/src/libraries/Microsoft.Extensions.Hosting/src/HostApplicationBuilder.cs#L144).

###Edit
Want to emphasize that I don't have a problem with the code or even merging the PR.  Especially since, as of the next release, it will just be a fallback value.

Reduced section header size.

twsouthwick:
> I see what you're doing, but am not sure it's the best approach. Especially since it's still relying on other assumptions. Ones which I know would fail for my company's sites.

Which assumptions would fail for your sites?

EmperorArthur:
Re-looking at this.  I believe I was a bit too hasty, and **the code looks good to me**.  While I it would not meet my needs, replicating aspnetcore behavior is the right call.

> Which assumptions would fail for your sites?

My case is a bit unique in I need two different projects in the same solution to have the same AppName. or at least override it for a `DataProtectionProvider`.  See #682.

twsouthwick:
oh yeah - i think you're use case is actually fairly common so that you can use the same data protection provider. this is just a default to match asp.net core :)

PR #675, opened 2026-04-08
Build whole solution and setup for copilot

twsouthwick:
We were only building a subset of projects due to CI issues with windows. This is to get everything building (and tests running) as well as configuring the host for copilot coding agent


CZEMacLeod:
Has there been any movement on getting the WebApplication.targets and dlls released as a separate package or licenced for inclusion in the SDK?
https://github.com/CZEMacLeod/MSBuild.SDK.SystemWeb/pull/89#issuecomment-3481719566

twsouthwick:
> Has there been any movement on getting the WebApplication.targets and dlls released as a separate package or licenced for inclusion in the SDK? [CZEMacLeod/MSBuild.SDK.SystemWeb#89 (comment)](https://github.com/CZEMacLeod/MSBuild.SDK.SystemWeb/pull/89#issuecomment-3481719566)

sorry that stalled out - I'll get that going again

src/Aspire.Hosting.IncrementalMigration/IIS/IISExpressLaunchDetailsExtensions.cs
```diff
@@ -219,9 +219,9 @@ private static bool TryParseOldStyleProject(string path, [MaybeNullWhen(false)]
 
         var propertyGroups = project
             .Descendants(MsbuildNS + "PropertyGroup");
-        var use64bitIISExpress = propertyGroups
+        var use64bitIISExpress = !(propertyGroups
             .Descendants(MsbuildNS + "Use64BitIISExpress")
-            .FirstOrDefault() is { } b && bool.TryParse(b.Value, out var use64BitValue) ? use64BitValue : true;
+            .FirstOrDefault() is { } b) || !bool.TryParse(b.Value, out var use64BitValue) || use64BitValue;
```
CZEMacLeod: This doesn't look right to me...
If there is no Use64BitISSExpress object (b) you then try and access the Value property?

twsouthwick: this was from #664 from some new warnings. It switched the .FirstOrDefault() to a negation and the `&&` to `||` (I think that's de morgan's law) so the right operand is never invoked if there is not value. However, at the time I found it very unreadable but merged anyway. With your confusion I rewrote it to be more readable but not in the previous form as that apparently isn't "simplified" (I could probably turn the warning off, but generally like it and I think my rewrite is more readable anyway)

CZEMacLeod: Yeah - that the FirstOrDefault was negated was really not clear - on further review, it looks like the code would have worked as-is, but I much prefer your rewrite as it is much clearer and cleaner about what is happening, not just on that line, but for all the properties.

twsouthwick:
splitting up into smaller prs

